PR #83, opened 2014-06-23
Allow overriding credentials from Configuration when calling login

jdart1:
(in particular, account ID & role)


iloveitaly:
This has been implemented in the login call https://github.com/NetSweet/netsuite/blob/4056d8d1c40007735c4e72526e57b5f43b58644c/lib/netsuite/actions/login.rb

PR #134, opened 2014-10-22
Fix Custom Field from mutating input attributes

greggroth:
I ran into an unexpected behavior because `CustomFieldList` and `CustomField` mutate the input attributes.  If you set a hash of attributes to a variable, and use that variable in an iteration to initialize new records, the second time around it was coming back with no value.  I realized this is because `CustomFieldList` deletes the `:value` key.

While I was in there,  I also made it so a `CustomFieldList` can be initialized with a collection of `CustomField` records.

I realize now that the pattern we're using in the initialize method is also mutating the input hash:

``` ruby
      def initialize(attributes = {})
        @internal_id = attributes.delete(:internal_id) || attributes.delete(:@internal_id)
        @external_id = attributes.delete(:external_id) || attributes.delete(:@external_id)
        initialize_from_attributes_hash(attributes)
      end
```

I'll get around to fixing this in the future.


greggroth:
Wow.  This PR is over 2 years old now.  I merged master back in, so it's good to go if you're 👍 .

iloveitaly:
@greggroth Whoa. This is super old. Looks great, merging in!

iloveitaly:
Merged with be66024...189cc68

PR #152, opened 2015-01-20
Credential fix for search

jdfrens:
If credentials are not specified in a `NetSuite.config` block, then `#search` actions fail.

Since `#search` passes its credentials into `NetSuite::Configuration.connection` as a SOAP header, it doesn't bother to pass the credentials to `.connection`.  Ultimately, `auth_header({})` is called which tries to find the credentials in the default configuration.  Since they aren't set, `#email` or `#password` (whichever one is called first) raises an error.

My solution is to pass the credentials from `#search` into `#connection` anyway.  This seemed to be the most straightforward to me.  `#auth_header` was already being called unnecessarily, so this solution isn't any less efficient.

### DIFF
--- a/lib/netsuite/actions/search.rb
+++ b/lib/netsuite/actions/search.rb
@@ -30,7 +30,7 @@ module NetSuite
         )
 
         NetSuite::Configuration
-          .connection(soap_header: preferences)
+          .connection({ soap_header: preferences }, credentials)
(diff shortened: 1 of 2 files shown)


iloveitaly:
Great, thanks! Merged with 5f9a01a

I think there is some cleanup in `NetSuite::Configuration` that could make this a bit cleaner. For now, this is perfect.

PR #153, opened 2015-01-20
Add vendor record

jdfrens:
Adds a `NetSuite::Records::Vendor` model class.

We are using this in our app, and it's working fine.


iloveitaly:
Thanks! Merged with 511e7a8

PR #155, opened 2015-02-11
Move incorrectly placed customer fields into record refs

greggroth:
They're actually record refs and not fields.

https://system.netsuite.com/help/helpcenter/en_US/srbrowser/Browser2014_2/schema/record/customer.html


iloveitaly:
Great! Merged with 4abfc79

PR #162, opened 2015-02-24
Add parent as record ref to customer

jdfrens:
The parent of a Records::Customer is official a record ref.


iloveitaly:
Merged with 7469b5b95aacf4189877a6290bf8c6275dce4b63

PR #163, opened 2015-02-24
Add employee information

tabfugnic:
Add necessary fields so an employee can be retrieved and created properly


iloveitaly:
Great, thanks! Merged with 62534b5

I had to make a couple modifications for it to work with other Employee updates & ruby < 2.0

PR #165, opened 2015-02-26
Job changes: search and parent as record ref.

jdfrens:
This PR:

* Moves `:parent` from the field list to the record ref list.
* Adds `:search` to the actions.

### DIFF
--- a/spec/netsuite/records/job_spec.rb
+++ b/spec/netsuite/records/job_spec.rb
@@ -200,6 +200,10 @@ describe NetSuite::Records::Job do
     end
   end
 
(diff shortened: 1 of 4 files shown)


spec/netsuite/records/job_spec.rb
```diff
@@ -200,6 +200,10 @@
     end
   end
 
+  describe '.search' do
+    it 'searches'
```
iloveitaly: Would be great to have some light specs here in the future :)

iloveitaly:
Thanks! Merged with c3254d5

PR #166, opened 2015-02-26
More fields on employee

jdfrens:
Adds these fields:

* `give_access`
* `is_sales_rep`
* `klass`
* `middle_name`
* `send_email`

### DIFF
--- a/lib/netsuite/records/employee.rb
+++ b/lib/netsuite/records/employee.rb
@@ -1,7 +1,6 @@
 module NetSuite
   module Records
     class Employee < Support::Base
-      include Support::Fields
       include Support::Actions
       include Support::RecordRefs
       include Namespaces::ListEmp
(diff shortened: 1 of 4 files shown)


lib/netsuite/records/employee.rb
```diff
@@ -1,20 +1,23 @@
 module NetSuite
   module Records
-
     class Employee < Support::Base
+      include Support::Fields
```
iloveitaly: `Fields` is already included in `Support::Base`

https://github.com/NetSweet/netsuite/blob/b0d659067335fc296ad9ab96f2e5458d4dfeae43/lib/netsuite/support/base.rb

Was there a reason you needed to include `Support::Fields` explicitly in this model?

jdfrens: I ran a quick test through my app, and this include doesn't seem to be necessary.  I'll take it out.

iloveitaly:
Great! Merged with 4949080539d619eaa1cdd6b495fb4c980ad9037d

PR #167, opened 2015-03-02
Equality and better metaprogramming for RecordRef.

jdfrens:
Improvements to `NetSuite::Records::RecordRef`.

Improved metaprogramming:
- Implements `#respond_to_missing?` (to complement `#method_missing`).

Added equality:
- Must have same type.
- Must have equal non-nil internal ids.
- Internal ids are converted to strings before they are compared.
- All other fields and attributes are ignored.
- `#hash` is overridden to hash on the string version of the internal id.


iloveitaly:
My instinct here is to leave the equality operators working as is. This seams to break the principal of least surprise: ActiveRecord, and other DB abstractions I've worked with, don't operate like this; I'm hesitant to break convention with the gems and systems I've worked with in the past.

Any thoughts here @greggroth?

iloveitaly:
> Implements #respond_to_missing? (to complement #method_missing).

Could you explain why you needed to implement this? My meta programming knowledge might be a bit lacking here.

Thanks for the PRs! Keep them coming!

jdfrens:
Actually, the definition of `==` here matches up with the one for ActiveRecord (see: http://stackoverflow.com/questions/4738439/how-to-test-for-activerecord-object-equality).

I just now found [a blog post](http://domon.cc/2012/05/25/equality-of-active-record-objects/) which suggests a few changes to my definition (like calling `super`).  I'd be happy to make this change.

jdfrens:
If you don't define `#respond_to_missing?`, then `#respond_to?` in inconsistent with the state of the object and `#method` won't work at all.

```ruby
# WITHOUT respond_to_missing?
[1] (pry) main: 0> rr = NetSuite::Records::RecordRef.new(foo: 5)
#<NetSuite::Records::RecordRef:0x007fca15b280b8 @internal_id=nil, @external_id=nil, @type=nil, @attributes={:foo=>5}>
[2] (pry) main: 0> rr.foo
5
[3] (pry) main: 0> rr.respond_to?(:foo)
false
[4] (pry) main: 0> rr.method(:foo)
NameError: undefined method `foo' for class `NetSuite::Records::RecordRef'
from (pry):4:in `method'
```

Honestly, I have to need for this right now, but I wanted equality for my specs.  Since I was in this file anyway, I thought I'd add `#respond_to_missing?`.

lib/netsuite/records/record_ref.rb
```diff
@@ -18,13 +18,25 @@ def initialize(attributes_or_record = {})
...
           @external_id = record.external_id if record.respond_to?(:external_id)
           @type        = record.class.to_s.split('::').last.lower_camelcase
         end
       end
 
+      def ==(other)
+        other.class == self.class &&
+          other.internal_id != nil && self.internal_id != nil &&
+          other.internal_id.to_s == self.internal_id.to_s
+      end
```
greggroth: This seems reasonable to me

drewish:
Looks like there's a year old PR to add `respond_to_missing?`: https://github.com/NetSweet/netsuite/pull/73

PR #168, opened 2015-03-04
Added VendorBill and VendorPayment (and related records)

prburke:
Added the following records (with tests):

+ VendorBill
+ VendorBillItem
+ VendorBillItemList
+ VendorBillExpense
+ VendorBillExpenseList
+ VendorPayment
+ VendorPaymentApply
+ VendorPaymentApplyList


lib/netsuite/records/vendor_bill_item_list.rb
```diff
@@ -0,0 +1,32 @@
+module NetSuite
+  module Records
+    class VendorBillItemList
```
iloveitaly: This is perfect for now, but it would be great to use the Sublist implementation to DRY some of the list code up going forward:

https://github.com/NetSweet/netsuite/blob/master/lib/netsuite/support/sublist.rb

Ex:

https://github.com/NetSweet/netsuite/blob/c29cffe0a13326e0e2283f29e516a8ca5f588b72/lib/netsuite/records/contact_access_roles_list.rb

lib/netsuite/support/actions.rb
```diff
@@ -28,8 +28,6 @@ def action(name)
             self.send(:include, NetSuite::Actions::GetSelectValue::Support)
           when :search
             self.send(:include, NetSuite::Actions::Search::Support)
-          when :search_more_with_id
```
iloveitaly: Good catch! This should have been removed long ago!

iloveitaly:
Great! Merged with 5ea7c9f

PR #173, opened 2015-03-20
Add 'Array.wrap' core extension for deleteList action

jcroteau:
- Adds ```Array.wrap``` core extension adopted from Rails::ActiveSupport
- Resolves exception when deleteList action is used for a single record


iloveitaly:
In a rails app, wouldn't this override an existing `Array.wrap` method created by rails since `netsuite` would be loaded after rails?

I'm hesitant to pull in a convenience method when it's only used once – I don't think the cost of maintaining the additional convenience method is worth the benefit. What are your thoughts?

jcroteau:
Good points. I'll just demote this then to a method within the deleteList action.

PR #174, opened 2015-03-25
Add the "memo" field to cash refunds. 

jasdeepgosal:
Perry straight forward, this was just a CashRefund field we need to be able to set.


iloveitaly:
Great! Merged with 163a09d

PR #175, opened 2015-03-31
Prevent exception when using deleteList on a single record

jcroteau:


lib/netsuite/actions/delete_list.rb
```diff
@@ -58,7 +58,17 @@ def request_body
...
       def response_list
-        @response_list ||= @response.to_hash[:delete_list_response][:write_response_list][:write_response]
+        @response_list ||= array_wrap(@response.to_hash[:delete_list_response][:write_response_list][:write_response])
+      end
+
+      def array_wrap(object)
+        if object.nil?
+          []
+        elsif object.respond_to?(:to_ary)
+          object.to_ary || [object]
```
greggroth: Do you have an example of when `object.to_ary` would return something falsey?

jcroteau: I just copied this from the ```Array.wrap``` extension provided by Rails, so it's meant for a much more general use case than it is here. It's possible, though perhaps not probable, that an object can have ```to_ary``` defined to return ```nil```. I think the intention of this method as written is to guarantee that the return value is an Array under all circumstances, including this one.

iloveitaly:
@jcroteau Thanks for your work here! I ended up adding a `array_wrap` to `Support::Requests` and fixing this bug for `get_all` and `delete_list`

2874194...528007b

jcroteau:
@iloveitaly glad to see it came in handy :)

PR #181, opened 2015-05-12
Add 2014_2 WSDL

adambedford:
Adding the most recent WSDL.


iloveitaly:
We've decided to stop including the WSDL files in the gem. Just reference the WSDL URL from here on out.

PR #184, opened 2015-05-20
Add TimeBill and TranEmp

adambedford:
This adds support for TimeBill record type and adds Employee transaction type.

This is now behaving as expected.


iloveitaly:
Merged this with 64e8a17...17de52f

I'm not convinced that the specs that we have the majority of records really add that much value. I think the NS gem tests should confirm that the NS gem logic is functioning correctly, but I don't think it makes sense for the test to answer the question:

1. Will this record operate properly in my NS instance? Every instance is different: it's up the gem user to fix any issues specific to their instance.
2. Are the fields configured correcly based on the schema? No need to duplicate the field list in the spec to ensure that fields configuration is correctly. If the field was setup as a `read_only_field` instead of a `field`, then the spec is most likely wrong as the developer misunderstood the schema/made a typo.

I've created a [basic record spec](https://github.com/NetSweet/netsuite/blob/master/spec/netsuite/records/basic_record_spec.rb) to try to create a generic test that we can easily add to new records that we create. Thoughts @adambedford?

PR #186, opened 2015-05-22
Add JobStatus

adambedford:
This adds support for JobStatus records.

I will try to write some specs sooner rather than later.


iloveitaly:
Merged with 64e8a17...17de52f

Issue #190, opened 2015-06-05
Anyone able to use Subsidiary Lists?


drewish:
I'm trying to specify a subsidiary list when creating a `ServiceSaleItem` using the syntax I found in [assembly_item_spec.rb](https://github.com/NetSweet/netsuite/blob/master/spec/netsuite/records/assembly_item_spec.rb#L26-L31):
```ruby
      subsidiary_list: {
        record_ref: [
          { internal_id: 1 },
        ]
      },
```

The XML it's generating for the upsert looks like:
```xml
<platformMsgs:upsert>
  <platformMsgs:record xsi:type="listAcct:ServiceSaleItem" platformMsgs:externalId="gold">
    <listAcct:subsidiaryList>
      <platformCore:recordRef>
        <platformMsgs:internalId>1</platformMsgs:internalId>
      </platformCore:recordRef>
    </listAcct:subsidiaryList>
    <listAcct:itemId>gold</listAcct:itemId>
    <listAcct:displayName>Gold plan</listAcct:displayName>
    <listAcct:salesDescription xsi:nil="true"/>
    <listAcct:taxSchedule internalId="1"/>
    <listAcct:isTaxable>true</listAcct:isTaxable>
  </platformMsgs:record>
</platformMsgs:upsert>
```
 
But I get the following error:
```xml
<soapenv:Envelope xmlns:soapenv="http://schemas.xmlsoap.org/soap/envelope/" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <soapenv:Body>
    <soapenv:Fault>
      <faultcode>soapenv:Server.userException</faultcode>
      <faultstring>org.xml.sax.SAXException: Expected {urn:core_2013_2.platform.webservices.netsuite.com}internalId, found {urn:messages_2013_2.platform.webservices.netsuite.com}internalId</faultstring>
      <detail>
        <ns1:hostname xmlns:ns1="http://xml.apache.org/axis/">partners-java10001.bos.netledger.com</ns1:hostname>
      </detail>
    </soapenv:Fault>
  </soapenv:Body>
</soapenv:Envelope>
```

Any clues what I'm doing wrong? I'd tried it with the 2011_2 and 2012_1 APIs as well with the same error.

When I fetch an existing `InventoryItem` with a subsidiary set the XML looks like:
```xml
[...]

PR #191, opened 2015-06-10
Specify the appropriate klass for subsidiary_list fields

drewish:
Solves https://github.com/NetSweet/netsuite/issues/190 by following the pattern from:
https://github.com/NetSweet/netsuite/blob/master/lib/netsuite/records/inventory_item.rb#L61
https://github.com/NetSweet/netsuite/blob/master/lib/netsuite/records/assembly_item.rb#L47

It fixed my issue with `ServiceSaleItem` so I'm assuming it's the right fix for the other classes as well, but I didn't actually test out creating them yet.

### DIFF
--- a/lib/netsuite/records/non_inventory_sale_item.rb
+++ b/lib/netsuite/records/non_inventory_sale_item.rb
@@ -28,12 +28,13 @@ module NetSuite
       record_refs :billing_schedule, :cost_category, :custom_form, :deferred_revenue_account, :department, :income_account,
         :issue_product, :item_options_list, :klass, :location, :parent, :pricing_group, :purchase_tax_code,
         :quantity_pricing_schedule, :rev_rec_schedule, :sale_unit, :sales_tax_code, :ship_package, :store_display_image,
-        :store_display_thumbnail, :store_item_template, :subsidiary_list, :tax_schedule, :units_type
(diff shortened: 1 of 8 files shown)


lib/netsuite/records/non_inventory_sale_item.rb
```diff
@@ -28,12 +28,13 @@ class NonInventorySaleItem
       record_refs :billing_schedule, :cost_category, :custom_form, :deferred_revenue_account, :department, :income_account,
         :issue_product, :item_options_list, :klass, :location, :parent, :pricing_group, :purchase_tax_code,
         :quantity_pricing_schedule, :rev_rec_schedule, :sale_unit, :sales_tax_code, :ship_package, :store_display_image,
-        :store_display_thumbnail, :store_item_template, :subsidiary_list, :tax_schedule, :units_type
+        :store_display_thumbnail, :store_item_template, :tax_schedule, :units_type
 
-      field :pricing_matrix, PricingMatrix
```
drewish: This was included twice. I dropped the first one so they'd be sorted alphabetically.

iloveitaly:
Great work! Keep em' coming!

Merged with b6e6ebd86008ba7142fda1d65525f4e4c1544f64

PR #192, opened 2015-06-11
Switch from should to expect

drewish:
Use the new syntax to avoid the should deprecation warnings.


spec/netsuite/records/credit_memo_apply_list_spec.rb
```diff
@@ -21,7 +21,7 @@
         }]
       }
 
-      expect(list.to_record.should).to eq(record)
```
drewish: It seemed like this didn't get removed when it was converted.

iloveitaly:
:+1: 

Merged with a4650b11a695fbc06ed4b99608ce05a43ca29bd7

PR #193, opened 2015-06-11
Invoice's subsidiary should be a record_ref

drewish:
Everywhere else subsidiary is a `record_ref` so just seems like this one got missed.


iloveitaly:
:+1: 

Merged with 56a123eb75e3ab6d496686cba89c0a5fec1f8dd2

PR #195, opened 2015-06-18
Since we advertise CRUD let's show off the D

drewish:
Add a simple example of the deletion. I'd expected it to be a class method, wasn't planning on needing to do a `.get(id).delete`.


drewish:
What do you think about adding `delete` as a class method? It doesn't seem like you should need to fetch the record before deleting it.

iloveitaly:
Nice catch! Merged with dc3c50c...46f1f09

PR #200, opened 2015-06-30
Added a basic spec & the to_be_printed flag to CashSale

jasdeepgosal:
CashSale had no specs and we needed to add a field so we added some basic specs and added the field we needed. 

In adding this spec, we discovered that the `exp_cost_discount` field was being overridden by the `record_ref` so we dropped it.


iloveitaly:
Awesome, thanks! Merged with d2af7e8...dc3c50c

PR #203, opened 2015-07-07
Correcting duplicate ISO Code

lorenpaulsen:
The code 'GU' appears in the list twice. The proper code for Equatorial Guinea is GQ instead of GU.


iloveitaly:
Great! Merged with 42ea61c...d2af7e8

PR #208, opened 2015-08-11
Add handling for WSDL 2013.2 CustomField internalId-scriptId transition

jcroteau:
Fix for issues #196 & #207 

Apparently, at NS SuiteTalk WSDL version 2013.2, CustomField ```internalId``` values were migrated to ```scriptId```s and the ```internalId``` values replaced with generated numeric ids. This was breaking CustomFieldList adds and updates when using WSDL 2013.2 and beyond.  The changes contained in this PR should address this problem.

PR #210, opened 2015-08-20
Updated the CustomerAddressBook fields to make it work

cfactolerin:
Created a new record called Address and updated the CustomerAddressBook fields to make it work.

### DIFF
--- a/lib/netsuite.rb
+++ b/lib/netsuite.rb
@@ -63,6 +63,7 @@ module NetSuite
     autoload :AssemblyItem,                     'netsuite/records/assembly_item'
     autoload :Account,                          'netsuite/records/account'
     autoload :AccountingPeriod,                 'netsuite/records/accounting_period'
+    autoload :Address,                          'netsuite/records/address'
     autoload :BaseRefList,                      'netsuite/records/base_ref_list'
     autoload :BillAddress,                      'netsuite/records/bill_address'
     autoload :BillingSchedule,                  'netsuite/records/billing_schedule'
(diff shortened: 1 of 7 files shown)


iloveitaly:
@cfactolerin Thanks for the PR! What exactly was breaking with the existing implementation? It's working fine for me in a couple projects.

cfactolerin:
Hi @iloveitaly. We encountered some errors when we were adding an Address to a Customer and when I looked at the latest schema, it's a bit different. https://system.netsuite.com/help/helpcenter/en_US/srbrowser/Browser2015_1/schema/other/customeraddressbook.html?mode=package

I'll try to see if I can reproduce the errors we saw and post it here. It returns an error if we create the CustomerAddressbook together with the Customer saying that CustomerAddressbook does not have fields that are in the Address object. If we create the Customer in advance and then update it with CustomerAddressbook data it will return true but somehow the Address is not added if get the customer again or in the UI.

I agree with you that this might break the existing codes you have since I removed some fields in CustomerAddressBook.

iloveitaly:
Got it! Thanks for the explination!

Given that it will break older versions, do you think you could conditionally enable the new syntax based on the `NetSuite::Configuration.api_version` similar to this PR https://github.com/NetSweet/netsuite/pull/208

Unfortunately, without this backwards compatibility it might be a while until I can merge this in :(

cfactolerin:
oooh so that's how you guys do it. Didn't know you can specify the version. I'll try to change it once I have the time. Sorry bout that.

PR #211, opened 2015-08-25
Adding serbia country code

nityaoberoi:
https://en.wikipedia.org/wiki/ISO_3166-2:CS


iloveitaly:
Great! Merged with 1ad7ee4...66a4698

PR #216, opened 2015-10-05
Opportunity record

iloveitaly:


lib/netsuite/records/opportunity.rb
```diff
@@ -0,0 +1,49 @@
...
+      attr_accessor :external_id
+      attr_accessor :search_joins
+
+      def initialize(attributes = {})
+        @internal_id = attributes.delete(:internal_id) || attributes.delete(:@internal_id)
+        @external_id = attributes.delete(:external_id) || attributes.delete(:@external_id)
+        initialize_from_attributes_hash(attributes)
+      end
+
+      def self.search_class_name
```
iloveitaly: No need for a `search_Class_name` here... this is auto-generated. This is only required when the search name is different from the class name.

lib/netsuite/records/opportunity_item_list.rb
```diff
@@ -0,0 +1,31 @@
+module NetSuite
+  module Records
+    class OpportunityItemList
```
iloveitaly: This should inherit from `Support::Sublist`. Checkout this sublist as an example:

https://github.com/NetSweet/netsuite/blob/master/lib/netsuite/records/invoice_item_list.rb

netsuite.gemspec
```diff
@@ -15,7 +15,7 @@ Gem::Specification.new do |gem|
   gem.require_paths = ['lib']
   gem.version       = Netsuite::VERSION
 
-  gem.add_dependency 'savon', '~> 2.3.0'
+  gem.add_dependency 'savon', '>= 2.3.0'
```
iloveitaly: There is a PR for this, let's leave this out of this PR if possible.

https://github.com/NetSweet/netsuite/pull/172

lib/netsuite/records/opportunity_item_list.rb
```diff
@@ -0,0 +1,30 @@
+module NetSuite
+  module Records
+    class OpportunityItemList < Support::Sublist
```
iloveitaly: @jheth `OpportunityItemList` shouldn't contain any logic aside a namespace definition and a `sublist` call. Ex: https://github.com/NetSweet/netsuite/blob/de60d36403587b9d255b9d85ec1d4651af0281cc/lib/netsuite/records/invoice_item_list.rb#L4-L6

iloveitaly:
Looking good! One additional comment: any new records should be added to the basic record spec:

https://github.com/NetSweet/netsuite/blob/master/spec/netsuite/records/basic_record_spec.rb

jheth:
Thanks @iloveitaly.  Just added to basic_record_spec.  Should I add an opportunity_spec.rb as well?

iloveitaly:
> Should I add an opportunity_spec.rb as well?

I've been bearish on individual specs for records. I'm fine with just a `basic_record_spec` entry.

If there is no unique record functionality, they add a lot of maintinance weight with very little value. What do you think?

Thanks for the PR!

iloveitaly:
Merged with https://github.com/NetSweet/netsuite/commit/371baf96794c5209357f6f5e2100868f000fc9c9

PR #228, opened 2015-11-20
Update journal_entry.rb

roeyby:
Fix incorrect namespace for JournalEntry (should be TranGeneral, not TranSales)

### DIFF
--- a/lib/netsuite/records/journal_entry.rb
+++ b/lib/netsuite/records/journal_entry.rb
@@ -5,7 +5,7 @@ module NetSuite
       include Support::RecordRefs
       include Support::Records
       include Support::Actions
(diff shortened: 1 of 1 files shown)


iloveitaly:
Merged with e4c1e8c...9fb9264. Thanks!

For any future PRs, making sure the tests pass would be much appreciated. Take a look at what I had to do for this PR: https://github.com/NetSweet/netsuite/commit/9fb92648f0f97b60c994225d03ae2c94c874685f

roeyby:
Michael,

I just added another PR so that both adds and searches works properly (they
are in different namespaces). I also verified that "bundle exec rspec" runs
with no errors.

Roey

PR #229, opened 2015-11-21
Add search namespace override to JournalEntry

roeyby:
The namespace for JournalEntry is tranGeneral, but it's search falls under the Transaction schema with the namespace tranSales. Add a search_class_namespace override to make this work properly.


iloveitaly:
Great, merged with 2b89c41

PR #235, opened 2016-01-05
Token-based authentication

bentoncreation:
What do you think about this for allowing token-based auth?

Instead of setting email, password, and role in the config, you add: consumer_key, consumer_secret, token_id, and token_secret.

(Note, you can't include ApplicationInfo when doing token-based auth--you'll get an error. I think it would be better if application_id was set in the config, and it included that in the header only when doing token-based auth.)

I've tested it with both user auth and token auth and it's working for me.

I'm not super happy with the way it chooses auth methods (configuration.rb:102)...any thoughts on how to make that better? Maybe it's better to have another config option for auth_type?

Also, I'm not sure the best way to test auth header hash, since the nonce, timestamp and signature can't be hard-coded (configuration_spec.rb:86).

### DIFF
--- a/lib/netsuite.rb
+++ b/lib/netsuite.rb
@@ -203,6 +203,11 @@ module NetSuite
     autoload :WorkOrderItemList,                'netsuite/records/work_order_item_list'
   end
 
+  module Passports
+    autoload :User,  'netsuite/passports/user'
+    autoload :Token, 'netsuite/passports/token'
+  end
+
   def self.configure(&block)
     NetSuite::Configuration.instance_eval(&block)
   end
--- a/lib/netsuite/configuration.rb
+++ b/lib/netsuite/configuration.rb
@@ -99,14 +99,30 @@ module NetSuite
     end
 
     def auth_header(credentials={})
-      {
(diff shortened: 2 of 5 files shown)


iloveitaly:
:+1: This is awesome! I'll take a look soon and get it merged in. Thanks!

bentoncreation:
Great, thanks!

iloveitaly:
@bentoncreation really great work here, thanks for the PR!

> I think it would be better if application_id was set in the config

Agreed!

> any thoughts on how to make that better?

I agree that is seems a bit messy, but I don't think there is a clearly better approach. We could add a `token_auth` boolean config option, but I'm not sure how that would add any value either when using per-request credentials or global credentials. I think your approach is great for now!

> Also, I'm not sure the best way to test auth header hash, since the nonce, timestamp and signature can't be hard-coded (configuration_spec.rb:86).

An option here would be mocking `nonce` and `timestamp` and tested the generated values. I think the tests you wrote up are more than fine for now!

Great stuff! Merged with ffd1e67..e790129

PR #244, opened 2016-02-25
Support for Folder creation

jdedels:


lib/netsuite/records/folder.rb
```diff
@@ -0,0 +1,26 @@
+module NetSuite
+  module Records
+    class Folder
```
iloveitaly: Minor thing here: I've created a [BasicRecordSpec](https://github.com/NetSweet/netsuite/blob/master/spec/netsuite/records/basic_record_spec.rb) to cover some common tests for all records. Any new records should be added to the `basic_record_list`

iloveitaly:
:+1: Merged with ffd1e679fd0cc9207dea75fee84edfe45e3763c5

PR #262, opened 2016-06-23
Add support for PaymentItem records

kristyfrey:


kristyfrey:
Need to fix tests.  Will create new pull request when they are fixed.

PR #264, opened 2016-06-29
Add custom fields to Job (Project)

ceckert:


lib/netsuite/version.rb
```diff
@@ -1,3 +1,3 @@
 module Netsuite
-  VERSION = '0.6.7'
+  VERSION = '0.6.8'
```
iloveitaly: @ceckert could you remove the version change? otherwise this looks good

iloveitaly:
Merged with c599e60a5c6dd2bb9334ffb11ffaa43016468075

PR #268, opened 2016-07-06
Support for ItemMemberList and ItemGroup

kristyfrey:
With ItemMemberList, we can now update members of KitItems and ItemGroups


iloveitaly:
Awesome, @kristyfrey!

Merged with 1b9a7c6...ed0fcd2

Quick tip: if there isn't anything unique about a record you are implementing, checkout `basic_record_spec`. Adding the record to the list of records that are tested is an easy way to help avoid writing extensive tests and get nearly the same result.

PR #274, opened 2016-07-26
Issue 272 item vendor list

### LINKED ISSUE #272: itemVendorList doesn't show up in the InventoryItem object
Hi there :)

I'm receiving the following soap response for an inventory item:

```xml
  <soapenv:Header>
    <platformMsgs:documentInfo xmlns:platformMsgs="urn:messages_2016_1.platform.webservices.netsuite.com">
      <platformMsgs:nsId>WEBSERVICEXXXXXX</platformMsgs:nsId>
    </platformMsgs:documentInfo>
  </soapenv:Header>
  <soapenv:Body>
    <getResponse xmlns="urn:messages_2016_1.platform.webservices.netsuite.com">
      <readResponse>
        <platformCore:status xmlns:platformCore="urn:core_2016_1.platform.webservices.netsuite.com" isSuccess="true"/>

<!-- and then vendor information like this: -->

          <listAcct:itemVendorList>
            <listAcct:itemVendor>
              <listAcct:vendor xmlns:platformCore="urn:core_2016_1.platform.webservices.netsuite.com" internalId="20">
                <platformCore:name>Spring Water</platformCore:name>
              </listAcct:vendor>
              <listAcct:purchasePrice>16.14</listAcct:purchasePrice>
              <listAcct:preferredVendor>true</listAcct:preferredVendor>
            </listAcct:itemVendor>
          </listAcct:itemVendorList>

<!-- the rest of the response ... -->

```

but in the response object the mentioned vendor information is missing

rudyyazdi:
Hey @iloveitaly following to our conversation, here is the implementation using `Sublist`. Hope it's ok. If you have any further comments please let me know and I'll fix them ;)

If you think it's a good idea I guess we can make vendor be an instance of `Records::Vendor` as well I can submit another PR for that.

This will fix #272

Cheers,

### DIFF
--- a/lib/netsuite/records/item_vendor.rb
+++ b/lib/netsuite/records/item_vendor.rb
@@ -1,0 +1,7 @@
+module NetSuite
+  module Records
+    class ItemVendor
+
+    end
+  end
+end
--- a/lib/netsuite/records/item_vendor.rb
+++ b/lib/netsuite/records/item_vendor.rb
@@ -1,7 +1,18 @@
 module NetSuite
   module Records
     class ItemVendor
+      include Support::Fields
+      include Support::Records
+      include Namespaces::ListAcct
 
+      fields :vendor, :purchase_price, :preferred_vendor
+
+      attr_reader :internal_id
+
+      def initialize(attributes = {})
+        @internal_id = attributes.delete(:internal_id) || attributes.delete(:@internal_id)
+        initialize_from_attributes_hash(attributes)
+      end
     end
   end
 end
--- a/lib/netsuite/records/item_vendor.rb
+++ b/lib/netsuite/records/item_vendor.rb
@@ -7,10 +7,7 @@ module NetSuite
 
       fields :vendor, :purchase_price, :preferred_vendor
 
-      attr_reader :internal_id
-
       def initialize(attributes = {})
-        @internal_id = attributes.delete(:internal_id) || attributes.delete(:@internal_id)
         initialize_from_attributes_hash(attributes)
       end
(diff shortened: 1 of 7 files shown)


lib/netsuite/records/item_vendor.rb
```diff
@@ -0,0 +1,18 @@
...
+      include Support::Fields
+      include Support::Records
+      include Namespaces::ListAcct
+
+      fields :vendor, :purchase_price, :preferred_vendor
+
+      attr_reader :internal_id
+
+      def initialize(attributes = {})
+        @internal_id = attributes.delete(:internal_id) || attributes.delete(:@internal_id)
```
iloveitaly: @rudyyazdi does the ItemVendor have a `internalId`? I'm not seeing one in the schema, but sometimes NS will return XML elements that don't exist in the schema

https://system.netsuite.com/help/helpcenter/en_US/srbrowser/Browser2015_1/schema/other/itemvendor.html?mode=package

rudyyazdi: @iloveitaly yeah you're right, the internal_id belongs to the `vendor` not the `item_vendor`. I have fixed it up :)

Thanks

iloveitaly:
> If you think it's a good idea I guess we can make vendor be an instance of Records::Vendor as well I can submit another PR for that.

The way you've implemented this is correct. The vendors returned in a sublist have a different structure from the standard vendor record, so a separate class should be used.

Thanks for this PR! It's merged with bc833d2..e56c356

PR #278, opened 2016-08-04
Notes support

charger:


lib/netsuite/records/note.rb
```diff
@@ -0,0 +1,36 @@
+module NetSuite
+  module Records
+    class Note
+      OUTGOING = '_outgoing'
+      INCOMING = '_incoming'
```
iloveitaly: @charger what's the idea behind these constants? Is this something that the gem user will use?

iloveitaly:
@charger Thanks for this!

Could you remove the wsdl from the PR? I've stopped including the wsdls in the gem since the URLs in the wsdls can change depending on the data center URL that is used.

lib/netsuite/records/note.rb
charger: @iloveitaly Yes, gem user can use these constants. Maybe this is better than use magic strings.

iloveitaly: I'm definitely a fan of avoiding magic strings. However, right now we don't include magic strings in the gem for other aspects of the API. It's best to omit these from the gem for now until we create a generic solution for handling magic strings.

charger: done

iloveitaly:
Looks great @charger! Thanks for the PR, and for making those tweaks.

This is merged with 34d42a4...0b48a12

PR #280, opened 2016-08-10
Search joins 4 invoice

rudyyazdi:
Hey Michael,

I hope you're not sick of my PRs just yet 😜

I was getting an error when I used the basic column search for Invoices:

```
NoMethodError: 
undefined method `search_joins=' for #<NetSuite::Records::Invoice:0x007fb672481e78>
```

I wrote a little test for it as well, which I can put into a shared example for similar records if you think that's a good idea.

Thanks

### DIFF
--- a/lib/netsuite/records/invoice.rb
+++ b/lib/netsuite/records/invoice.rb
@@ -48,6 +48,7 @@ module NetSuite
 
       attr_reader   :internal_id
       attr_accessor :external_id
+      attr_accessor :search_joins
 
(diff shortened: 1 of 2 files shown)


iloveitaly:
> I hope you're not sick of my PRs just yet

Haha! Not at all. Thanks so much for your contributions! They are very appreciated 👍 

I like the idea about moving this test into a shared example. If there was a way we could run this test in `basic_spec` that would be awesome as well.

Merged with 21c59e4...34d42a4

PR #283, opened 2016-08-16
Refactor list objects to use Support::Sublist

davidlaprade:
It seemed there was common behavior in a bunch of the list objects could be refactored to use `Support::Sublist` instead. So I gave it a go.

Happy to combine all of this into a single commit, if that's preferable!

### DIFF
--- a/lib/netsuite/records/base_ref_list.rb
+++ b/lib/netsuite/records/base_ref_list.rb
@@ -2,31 +2,15 @@
 
 module NetSuite
   module Records
-    class BaseRefList
-      include Support::Fields
+    class BaseRefList < Support::Sublist
       include Support::Actions
       include Namespaces::PlatformCore
 
       actions :get_select_value
 
-      fields :base_ref
+      sublist :base_ref, RecordRef
 
-      def initialize(attrs = {})
-        initialize_from_attributes_hash(attrs)
-      end
-
-      def base_ref=(refs)
-        case refs
-        when Hash
-          self.base_ref << RecordRef.new(refs)
-        when Array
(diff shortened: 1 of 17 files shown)


lib/netsuite/records/base_ref_list.rb
```diff
@@ -2,31 +2,15 @@
 
 module NetSuite
   module Records
-    class BaseRefList
-      include Support::Fields
+    class BaseRefList < Support::Sublist
```
iloveitaly: @davidlaprade this class is a bit different than the others. It has the `get_select_value` action which can be called on this class directly. Could you try testing this against a live NS instance to see if it functions well? There's an example in the readme.

Otherwise, this PR looks great!!

davidlaprade: I tested it, and it looks like it's working for standard records:

```ruby
NetSuite::Records::BaseRefList.get_select_value(
  recordType: 'serviceSaleItem',
  field: 'taxSchedule'
)
```
image

And for custom records:

```ruby
NetSuite::Records::BaseRefList.get_select_value(
  field: 'custrecord_....',
  customRecordType: {
    '@internalId' => 10,
    '@xsi:type' => 'customRecord'
  }
)

```

image

Also, I should have noted this when I submitted, but I manually tested examples of almost all of the lists that I tweaked in this way to make sure they were still working. I think there were only one or two that I couldn't test. (I believe they were the BillingSchedule ones.)

davidlaprade: @iloveitaly ^^

iloveitaly:
@davidlaprade Merged with f85d2ed...fa853c0!

PR #284, opened 2016-08-18
Support pulling roles, WSDL, and account_ids using only email and password

davidlaprade:
@iloveitaly This is a work in progress/proof of concept. Just throwing it up here to get feedback on the approach, architecture, etc!

Usage:

```ruby
NetSuite::Rest::Utilities::Roles.get(
  email: NetSuite::Configuration.email,
  password: NetSuite::Configuration.password
)
=> {
 :accounts=>["TSTDRV15"],
 :roles=>[3, 1074, 1, 46],
 :wsdls=>["https://webservices.na1.netsuite.com"]
}
```

Still needs tests, and error handling.

### DIFF
--- a/lib/netsuite/rest/utilities/roles.rb
+++ b/lib/netsuite/rest/utilities/roles.rb
@@ -1,0 +1,85 @@
+require 'net/http'
+require 'uri'
+require 'json'
+
+module NetSuite
+  module Rest
+    module Utilities
+      class Roles
+
+        ROLES_API = "https://rest.netsuite.com/rest/roles"
+
+        def self.get(credentials)
+          new(credentials).get
+        end
+
+        def initialize(credentials)
+          @email     = encode_email(credentials.fetch(:email))
+          @signature = credentials.fetch(:password)
+        end
+
+        def get
+          response = make_request
+          parsed   = JSON.parse(response.body)
+          if reponse.code.to_i == 200
+            extract(parsed)
+          else
+            parsed
+          end
+        end
+
+        private
+
+        def encode_email(unencoded_email)
+          URI.escape(
+            unencoded_email,
+            Regexp.new("[^#{URI::PATTERN::UNRESERVED}]")
+          )
+        end
+
+        def make_request
+          uri = URI(ROLES_API)
+          request = Net::HTTP::Get.new(uri)
+          request['AUTHORIZATION'] = "NLAuth " +
+            "nlauth_email=#{@email},nlauth_signature=#{@signature}"
+
+          http = Net::HTTP.new(uri.hostname, uri.port)
+          http.use_ssl = true
+
+          http.start {|http| http.request(request)}
+        end
+
+        def extract(parsed)
+          # [
+          #   {
+          #    "account"=>{
+          #      "internalId"=>"TSTDRV15",
+          #      "name"=>"Honeycomb Mfg SDN (Leading)"},
+          #    "role"=>{
+          #      "internalId"=>3,
+          #      "name"=>"Administrator"},
+          #    "dataCenterURLs"=>{
+          #      "webservicesDomain"=>"https://webservices.na1.netsuite.com",
+          #      "restDomain"=>"https://rest.na1.netsuite.com",
+          #      "systemDomain"=>"https://system.na1.netsuite.com"}
+          #    },
+          #   {
+          #    "account"=>{
+          #      "internalId"=>"TSTDRV15",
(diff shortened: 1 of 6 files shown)


lib/netsuite/rest/utilities/roles.rb
```diff
@@ -0,0 +1,79 @@
...
+
+        def connection
+          Faraday.new(API) do |faraday|
+            faraday.request        :url_encoded
+            faraday.adapter        Faraday.default_adapter
+            faraday.authorization  :NLAuth,
+              nlauth_email:        @email,
+              nlauth_signature:    @signature
+          end
+        end
```
iloveitaly: @davidlaprade I love the idea of building this into the gem, but I'm hesitant to include a external gem dependency for a single HTTP request. Extra dependencies == extra headaches for developers trying to upgrade other gems.

Given the simplicity of this request, I'd prefer to use a ruby stdlib http client like `Net::HTTP`.

What do you think?

iloveitaly:
@davidlaprade nice find! I'm curious—what other endpoints are available under `rest.netsuite.com`?

davidlaprade:
@iloveitaly good call on `Net::HTTP`. Thoughts on this?

Not sure what's going on with CI. Tests all pass locally.

lib/netsuite/rest/utilities/roles.rb
```diff
@@ -0,0 +1,85 @@
...
+
+        private
+
+        def encode_email(unencoded_email)
+          URI.escape(
+            unencoded_email,
+            Regexp.new("[^#{URI::PATTERN::UNRESERVED}]")
+          )
```
iloveitaly: @davidlaprade why is this required?

lib/netsuite/rest/utilities/roles.rb
```diff
@@ -0,0 +1,85 @@
...
+
+        def make_request
+          uri = URI(ROLES_API)
+          request = Net::HTTP::Get.new(uri)
+          request['AUTHORIZATION'] = "NLAuth " +
+            "nlauth_email=#{@email},nlauth_signature=#{@signature}"
+
+          http = Net::HTTP.new(uri.hostname, uri.port)
+          http.use_ssl = true
+
```
iloveitaly: We should set the default timeout to something like 30 second

lib/netsuite/rest/utilities/roles.rb
```diff
@@ -0,0 +1,85 @@
...
+module NetSuite
+  module Rest
+    module Utilities
+      class Roles
+
+        ROLES_API = "https://rest.netsuite.com/rest/roles"
+
+        def self.get(credentials)
+          new(credentials).get
+        end
```
iloveitaly: @davidlaprade Nice use of `self.get` to make things cleaner for the gem user. Minor nit here: using `get(*args)` and `new(*args)` would make it easier to modify the underlying arguments and self-document that the class method is just a pass-through to the child method.

lib/netsuite/rest/utilities/roles.rb
davidlaprade: @iloveitaly 

> why is this required?

You mean the email encoding? I'm not sure why NetSuite requires it. But I couldn't get it to work without it.

Would this be more descriptive?

```ruby
URI.escape(email, /[#{URI::PATTERN::RESERVED}]/)
```

lib/netsuite/rest/utilities/roles.rb
davidlaprade: Sounds good

lib/netsuite/rest/utilities/roles.rb
davidlaprade: Sure!

davidlaprade:
@iloveitaly 

How would you feel about my building an extra layer of abstraction here? I'm thinking about a `NetSuite::Rest::Request` object, which would abstract away the details of actually making the API calls in the various `Rest` actions. Some nice things about the approach:
* we could centralize any application-level configurations (e.g. timeouts, protocols) on our http library so that we aren't repeating ourselves each time we write a new Rest action
* it'd be easy to reuse request object for other rest actions if we ever wanted to support them in the future
* it'd be super easy to swap out `Net::HTTP` for a full-blown HTTP client library if we ever wanted to

davidlaprade:
@iloveitaly review? Hopefully this is ready to go now

iloveitaly:
👍 

Merged with d0fff7b...b53cf50

PR #281, opened 2016-08-11
Record ref doesn't pick up the internal_id

rudyyazdi:
Hi Michael,

I was having a problem with record_ref and I figured it's happening because `RecordRef` expects the hash attribute to be something like this: `{internal_id: "44"}` but in the search result if you passed something like:

```
{
  :entity=>{:search_value=>{:@internal_id=>"11"}},
  :internal_id=>{:search_value=>{:@internal_id=>"5"}}
}

# it would turn it into:
{
  :entity=>"11",
  :internal_id=>"5"
}

# instead of this:
{
  :entity=>{:@internal_id=>"11"},
  :internal_id=>"5"}
}
```

So I added `k == :internal_id` in line 56. Hope my implementation is ok.

Cheers,

PR #286, opened 2016-08-18
refactored search_result.rb

rudyyazdi:
Hey Michael,

Here we go as per our discussion in #281 :)

Thanks


iloveitaly:
@rudyyazdi thanks so much for this PR! I don't have the bandwidth right now to dig into this, but I'll circle back as soon as I do!

PR #288, opened 2016-08-24
Adding upsert capability to Vendor

jeperkins4:
The update method does not appear to actually update the record on NetSuite.  I had to make several fields read only as well.

### DIFF
--- a/lib/netsuite/records/vendor.rb
+++ b/lib/netsuite/records/vendor.rb
@@ -7,7 +7,7 @@ module NetSuite
       include Support::Actions
       include Namespaces::ListRel
 
-      actions :get, :get_list, :add, :update, :delete, :search
+      actions :get, :get_list, :add, :update, :upsert, :delete, :search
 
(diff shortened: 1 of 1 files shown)


lib/netsuite/version.rb
```diff
@@ -1,3 +1,3 @@
 module Netsuite
-  VERSION = '0.7.1'
+  VERSION = '0.7.1.2'
```
iloveitaly: @jeperkins4 Can you remove the version bump? I'll handle this when I merge the branch.

iloveitaly:
@jeperkins4 Thanks for the PR! Do you mind squashing the changes into a single commit to keep the version history clean?

jeperkins4:
Mike, This should be a bit cleaner.  Thanks for the review and for considering this PR!

iloveitaly:
Perfect! Merged with 76f558f1a462949b64d527238d2a12dc35018603

PR #291, opened 2016-09-12
vendor credits also can be added and updated

rhfung-flexport:
Vendor credits also can be added and updated.


iloveitaly:
Awesome! Merged with 240e6b1120b0e3ca5bfd43782e8d2406732af331

PR #294, opened 2016-09-22
Fix bug in #get_all when the user lacks permission

davidlaprade:
Try `NetSuite::Records::Currency.get_all` with a role that lacks
Currency List permissions. The response you get is:

```XML
      <platformCore:code>INSUFFICIENT_PERMISSION</platformCore:code>
      <platformCore:message>Permission Violation: You need  the
        'Lists -&gt; Currency' permission to access this page. Please
        contact your account administrator.</platformCore:message>
    </platformCore:statusDetail>
  </platformCore:status>
```

Because of this[ `response_hash[:record_list]`](https://github.com/NetSweet/netsuite/blob/eb13f4b5c8d8eba8e0a7f4906eaebda455d641a3/lib/netsuite/actions/get_all.rb#L41) is `nil`, and so `response_hash[:record_list][:record]` raises:

```
  NoMethodError: undefined method `[]' for nil:NilClass
```

This prevents that from happening

### DIFF
--- a/lib/netsuite/actions/get_all.rb
+++ b/lib/netsuite/actions/get_all.rb
@@ -38,7 +38,11 @@ module NetSuite
       end
 
       def response_body
-        @response_body ||= array_wrap(response_hash[:record_list][:record])
+        @response_body ||= if response_hash[:record_list]
+          array_wrap(response_hash[:record_list][:record])
+        elsif response_hash[:status][:status_detail][:@type] == "ERROR"
+          NetSuite::Error.new(response_hash[:status][:status_detail])
+        end
       end
 
       def response_hash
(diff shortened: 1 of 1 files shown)


lib/netsuite/actions/get_all.rb
```diff
@@ -57,7 +61,7 @@ def get_all(credentials = {})
             if response.success?
               response.body.map { |attr| new(attr) }
             else
-              raise RecordNotFound, "#{self} with OPTIONS=#{options.inspect} could not be found"
+              raise RecordNotFound, response.body.message
```
iloveitaly: This approach is different from the class-level methods like `search` and `get_list`. Those methods return false if the operation is not successful instead of raising an exception. Also, `RecordNotFound` isn't the correct error here: users may be leveraging that error class for situations where the record doesn't actually exist in the NS instance.

The issue with that approach is the errors are not exposed to the user for class-level operations.

For now, let's keep with the current convention. Later on we can rework how class-level operations work and expose the errors to the user in a useful wayl.

iloveitaly:
I fixed this with ebd8176b71d286ad650d3e3919a92e01bbefa1d3

PR #295, opened 2016-10-04
Add missing #search_joins reader/writer

greggroth:
Without this, you get an exception when retrieving the results from a
saved search.


iloveitaly:
Thanks @greggroth!

PR #296, opened 2016-10-12
Allow usage of `type` to instantiate a custom field from a hash

bernardeli:
Hi,

I've been working on a netsuite integration and this gem fit like a glove to me. So many thanks for all work that has been put into it.

To make things simpler, I first build a hash with the appropriate keys and sub-hashes, and only then, I initialise a netsuite record to call `#add`.

I realised that a CustomField already takes [`type` or `@xsi:type`](https://github.com/NetSweet/netsuite/blob/master/lib/netsuite/records/custom_field.rb#L13) to initialise.

Since I use CustomFields initialised from within a CustomFieldList, I'd rather use type so my code looks like more idiomatic as opposed to `@xsi:type`. 

I've added a test to make sure the output is as expected.

Let me know if any changes are required.

Cheers
Ricardo

### DIFF
--- a/lib/netsuite/records/custom_field_list.rb
+++ b/lib/netsuite/records/custom_field_list.rb
@@ -97,7 +97,7 @@ module NetSuite
 
         def extract_custom_field(custom_field_data)
           # TODO this seems brittle, but might sufficient, watch out for this if something breaks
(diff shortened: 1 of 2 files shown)


iloveitaly:
@bernardeli Glad you've found the gem useful! Looks like a great change. Thanks.

PR #297, opened 2016-10-25
Handle more data types in custom field list

kristyfrey:


lib/netsuite/records/custom_field_list.rb
```diff
@@ -119,6 +119,13 @@ def create_custom_field(reference_id, field_value)
           when field_value.is_a?(FalseClass),
                field_value.is_a?(TrueClass)
             'BooleanCustomFieldRef'
+          when field_value.is_a?(String)
+            'StringCustomFieldRef'
+          when field_value.is_a?(Integer)
+            'LongCustomFieldRef'
```
iloveitaly: @kristyfrey did NS throw an error when `StringCustomFieldRef` and `LongCustomFieldRef` were not explicitly defined? I don't recall NetSuite strictly requiring these field types.

iloveitaly:
Closing this out. In my experience NS does not require you to define these types when submitting custom field lists of these values.

Issue #222, opened 2015-10-14
Cache Remote WSDL


iloveitaly:
* Right now, the WSDL content is pulled on _each_ Netsuite call!
* By default, savon will cache the WSDL content per-instance. Right now we create a new savon instance on each request. We should continue doing this as connection configuration can change on a per-request basis
* Savon offers the option of providing WSDL content instead of pulling it from a URL https://gist.github.com/rubiii/427832#file-savon_instantiation-rb-L42-L47
* We can grab the WSDL URL from the current configuration `NetSuite::Configuration.wsdl`
* The WSDL will be unique on `{ data center url, API version}`
* For starters, we can cache the string content of the WSDL in memory and clear on `NetSuite::Configuration.reset!`

PR #301, opened 2016-11-20
WSDL caching

davidlaprade:
Ping @iloveitaly. Review?

Noted relevance to #222. I'll take a deeper look at that approach soon


davidlaprade:
Noted relevance to https://github.com/NetSweet/netsuite/issues/222. I'll take a deeper look at that approach soon

iloveitaly:
@davidlaprade this looks good, although the naming is a bit off. What we are really doing here is caching the Account => DataCenter URL mapping. The WSDL URL is a simple string concatenation once we have the correct DS URL for the account.

When we finish up the work for #222 we should the naming convention here to represent that we are caching the account > DS URL mapping.

PR #305, opened 2016-11-29
Add `#update` action to `NetSuite::Records::NonInventorySaleItem#update`

kruppel:
In the event the external ID needs to be updated for a non-inventory sale item, an update action must be available.

We had this happen in our case and attempted to just use `#upsert`. However, since upserting depends on matching external IDs, we got back error responses.


iloveitaly:
Thanks for the PR @kruppel!

PR #308, opened 2016-12-20
Add getDeleted Action

greggroth:
The documentation on this is pretty light, so this was assembled by
using the WSDL directly as reference.


iloveitaly:
@greggroth you are the man! Thanks for this. Merged.

Will cut a fresh gem soon

PR #309, opened 2016-12-30
Cache missing record requests

davidlaprade:
Currently `NetSuite::Utilities#get_record` doesn't cache the results of requests for non-existent records, e.g. for record X when there is no record with id X. But it probably should. These requests will return the same response even if made multiple times.

Check out the spec to reproduce the issue this fixes.

The tests actually pass. CI is failing b/c of a caching issue with Ruby/nokogiri:

image


davidlaprade:
@iloveitaly I intended dd32795 to be optional.

iloveitaly:
@davidlaprade looks great, thanks David!

PR #320, opened 2017-02-20
WIP Thread safe per-request configuration

iloveitaly:
Passing a credentials hash doesn't work. There are too many configuration options (oauth, wsdl, api version, logging, etc).

Instead, we should allow the entire configuration object to be passed as an optional parameter

---

- [ ] investigate if there is any downside to removing namespaces
- [ ] rework wsdl cache into higher level abstraction

### DIFF
--- a/lib/netsuite.rb
+++ b/lib/netsuite.rb
@@ -253,6 +253,7 @@ module NetSuite
   end
 
   def self.configure(&block)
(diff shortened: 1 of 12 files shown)


iloveitaly:
superseded by https://github.com/NetSweet/netsuite/pull/549

PR #321, opened 2017-03-14
Remove errant `include`

greggroth:
`Support::Records` also implements `to_record`, which overrides the
implementation in `Support::Sublist`.  This causes the items on the
member list to be ommited in the request body.


iloveitaly:
Awesome!! Thanks. Merged.

PR #322, opened 2017-03-14
Update .ruby-version

greggroth:
According to the [ruby versions specified on CI](https://github.com/NetSweet/netsuite/blob/master/circle.yml#L5), this gem dropped support for 1.9.3.

And in any case, it makes sense to use a more modern ruby for development (or remove this file altogether).


iloveitaly:
👍 Great point! Merged.

PR #326, opened 2017-03-21
Add AssemblyBuild and AssemblyUnbuild record types

greggroth:


iloveitaly:
@greggroth can you add these new records to the basic spec list?

https://github.com/NetSweet/netsuite/blob/master/spec/netsuite/records/basic_record_spec.rb

greggroth:
Done!  Thanks for mentioning that spec.  It pointed out a bug I had (I used `:class` as a record ref when I needed to use `:klass`).

iloveitaly:
Awesome! Merged.

PR #328, opened 2017-03-23
Add support for CustomerDeposit apply list records

davidlaprade:
@iloveitaly: This should be pretty straightforward.


iloveitaly:
👍 merged

PR #340, opened 2017-07-07
Support class attribute in InventoryAdjustment

monkbroc:
Element InventoryAdjustment now has a class RecordRef. See https://webservices.netsuite.com/xsd/transactions/v2017_1_0/inventory.xsd on line 24

Without the `:klass` record ref, I get an error when loading an inventory adjustment record with API 2017_1.


monkbroc:
Thanks for your work maintaining this gem. It bring invaluable sanity to the world of NetSuite integrations! :heart:

Issue #343, opened 2017-07-17
Netsuite::Error should respond to "to_s" with something of note


Senjai:
![image](https://user-images.githubusercontent.com/1831235/28292534-43bcbeac-6b04-11e7-9dd1-b83e79ab9e04.png)

This should probably dump out the actual error when to_s is called on each error.

iloveitaly:
@Senjai good idea! Want to submit a PR?

PR #346, opened 2017-07-24
Allow errors to be represented as strings

Senjai:
Changes viewing errors in honeybadger from

`"#<NetSuite::Error:0x0055a605e61e90>"`

to

`"ERROR USER_ERROR You cannot edit the end of group line.  You must
delete the group."`

Which is helpful for debugging

Fixes: #343

### DIFF
--- a/lib/netsuite/errors.rb
+++ b/lib/netsuite/errors.rb
@@ -11,5 +11,7 @@ module NetSuite
       @code    = args[:code]
       @message = args[:message]
     end
+
+    alias_method :to_s, :inspect
   end
 end


Senjai:
There should probably be a string representation for all records. Even if it just delegates to inspect, this is a larger change than I wanted to make without additional input though.

iloveitaly:
@Senjai I like this, but I think the message should include some indication that the message is from a `NetSuite::Error` object. I believe this is a convention used by most API libraries, although I can't think of any great examples at the moment. I'll try to look into this over the next week.

Senjai:
@iloveitaly I can call super here which includes the object test and the object id as well. Would that suffice? I can then include the other data as tertiary.

Alternatively, I can delegate to inspect. Inspect includes a string representation of all instance variables.

Lmk

iloveitaly:
> Alternatively, I can delegate to inspect. Inspect includes a string representation of all instance variables.

I like this approach! Let's go with this.

Senjai:
@iloveitaly Bumped

```ruby
[3] pry(main)> NetSuite::Error.new.to_s
=> "#<NetSuite::Error:0x00558447e56a48 @type=nil, @code=nil, @message=nil>"
```

iloveitaly:
Awesome! Thanks @Senjai

PR #353, opened 2017-08-03
Fix VERSION constant module

jarednorman:
It was under `Netsuite` instead of `NetSuite` (capitalization) like
everything else in the gem. With this change there are no remaining references to `Netsuite`.

I noticed this while working on an application that kept its own NetSuite integration code under `Netsuite` to avoid collisions.


iloveitaly:
@jarednorman Nice catch! Merging this in.

PR #354, opened 2017-08-09
Add department to inventory adjustment item

monkbroc:
I added class to InventoryAdjustmentInventory in #340. This PR also adds department to that record type.


iloveitaly:
Awesome! Thanks @monkbroc

PR #364, opened 2017-10-12
add search to support case

myrovh:
As far as I've been able to test. Adding the search action to the Support Case record works fine with out any further modification needed.


iloveitaly:
Awesome! Thanks @myrovh

PR #366, opened 2017-10-25
Add a SalesOrder search example for "Pending Approval" SO's.

hdeshev:
As discussed on Slack -- adding an example for this type of SO search that shows how to use the  `TransactionStatus._salesOrderPendingApproval` value for the `status` field.

More valid values are available in the Schema browser: https://system.netsuite.com/help/helpcenter/en_US/srbrowser/Browser2016_2/schema/enum/transactionstatus.html

I'm not sure about just pasting the Schema browser URL in a comment like that. Perhaps I should just remove that line.

### DIFF
--- a/README.md
+++ b/README.md
@@ -404,6 +404,26 @@ NetSuite::Records::SalesOrder.search({
   }
 }).results
 
+# Search for SalesOrder records with a "Pending Approval" status using the TransactionStatus enum value.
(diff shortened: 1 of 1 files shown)


iloveitaly:
@hdeshev Awesome! Thanks so much. Schema link can't hurt anyone—we can always clean it up later when the docs are updated in a bigger way.

PR #370, opened 2017-11-16
Fix multipage search not passing credentials

avlazarov:
Performing a search action against a record class did not carry the explicitly passed credentials, which made the first page request and the following to have different authentication data.


iloveitaly:
This is awesome! Thanks!

PR #374, opened 2017-12-01
Add inter company journal entry

tatecarr:
Very similar to the regular JournalEntry but allows for from and to subsidiaries as well as setting the subsidiary on the line level.


iloveitaly:
Thanks @tatecarr!

PR #376, opened 2017-12-07
Add update list

AgustinSuarez:
This PR allows sending references to other record type when using the initialize call. For example SO to Item Fulfillment, you can specify which ship group you want NS to return to you. Thats done via an aux_reference.

My mistake I also committed all those new changes in the same branch as the Update List PR that was done a couple months ago. That code remains unchanged.


lib/netsuite/support/records.rb
```diff
@@ -35,6 +35,12 @@ def to_attributes!(hash, kname, v)
           hash[:attributes!][kname]['externalId'] = v.external_id
         end
 
+        if v.respond_to?(:replace_all) && !v.replace_all.to_s.empty?
+          hash[:attributes!] ||= {}
+          hash[:attributes!][kname] ||= {}
+          hash[:attributes!][kname]['replaceAll'] = v.replace_all
+        end
```
iloveitaly: @AgustinSuarez why did this functionality need to be moved out of the sublist module? I believe `replaceAll` is only applicable to sublists, so I don't think it should be required anywhere else. What am I missing here?

iloveitaly:
@AgustinSuarez This looks great! I just left one comment in the code. Two other items that came to mind:

* Can you rebase this to master?
* Would be awesome to get a `readme` entry with some usage examples for the community.

Thanks again!

lib/netsuite/support/records.rb
AgustinSuarez: So in order to upsert sales orders and other type of transaction correctly,  you need to be able to define it as follows  `<listRel:addressbookList replaceAll="false"/>` The problem with the previous code is that it added it as a field and didn't take it into consideration. 

This helps when upserting records like addresses and line items

AgustinSuarez:
Hey @iloveitaly It's already rebased on master. I did on my machine and there are no changes done. I have also added the new documentation

Thank you

AgustinSuarez:
@iloveitaly My mistake. Had an issue with git. I have rebased on master. Thank you!

PR #379, opened 2017-12-19
Fix Libya's NetSuite country value

hdeshev:
Use just `_libya` instead of `_libyanArabJamahiriya`

I discovered that one while syncing a Customer record with an address in Libya.

[Docs](https://system.netsuite.com/help/helpcenter/en_US/srbrowser/Browser2017_2/schema/enum/country.html?mode=package) for the enum value.


iloveitaly:
Nice catch! Merged.

Issue #382, opened 2018-01-16
Netsuite have changed how sandbox urls work


perryn:
Sandboxes no longer have a different url, but just use a different account id, and this means the 'sandbox' configuration option no longer works as documented. 

(A workaround for using this gem with a sandbox account is to set 'sandbox' to false and append _SB1 to the account id)

Further info at : https://netsuite.custhelp.com/app/answers/detail/a_id/66484

iloveitaly:
@perryn Thanks for calling this out! Want to submit a PR to fix this?

PR #383, opened 2018-01-26
Updates sandbox url to new address

j-mutter:
Updates the sandbox WSDL url to the new address.

I manually checked this with `api_version`s of both `2017_1` and `2015_1`

Closes https://github.com/NetSweet/netsuite/issues/382

@iloveitaly

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ NetSuite.configure do
   api_version	'2012_1'
 
   # optionally specify full wsdl URL (to switch to sandbox, for example)
-  wsdl          "https://webservices.sandbox.netsuite.com/wsdl/v#{api_version}_0/netsuite.wsdl"
+  wsdl          "https://system.netsuite.com/wsdl/v#{api_version}_0/netsuite.wsdl"
 
   # if your datacenter is being switched, you'll have to manually set your wsdl location
   wsdl          "https://webservices.na2.netsuite.com/wsdl/v#{api_version}_0/netsuite.wsdl"
--- a/lib/netsuite/configuration.rb
+++ b/lib/netsuite/configuration.rb
@@ -124,7 +124,7 @@ module NetSuite
       else
         attributes[:wsdl] ||= begin
           if sandbox
-            "https://webservices.sandbox.netsuite.com/wsdl/v#{api_version}_0/netsuite.wsdl"
+            "https://system.netsuite.com/wsdl/v#{api_version}_0/netsuite.wsdl"
           else
             "https://#{wsdl_domain}/wsdl/v#{api_version}_0/netsuite.wsdl"
           end
@@ -138,7 +138,7 @@ module NetSuite
       else
         # if sandbox, this parameter is ignored
         if sandbox
-          'webservices.sandbox.netsuite.com'
+          'system.netsuite.com'
         else
           attributes[:wsdl_domain] ||= 'webservices.netsuite.com'
         end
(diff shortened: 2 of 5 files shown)


tatecarr:
@j-mutter Should we use a flag to allow both the new approach and old approach to sandbox endpoints?  Maybe default to sandbox being the new endpoint style, but the flag will force old style still?  Thinking some people could want to update their gem version, but still have the old sandbox.  At this point there doesn't appear to be a hard cutover to the new style, so there could be a period where people want the latest gem but need to access both sandbox endpoints.  Then just have an update requirement for the next version of the gem to set that flag if you still need the old SB endpoint.

Or perhaps we update the gem to use the new style.  But we could add a similar config variable to allow for `wsdl_domain_sandbox` similar to being able to set `wsdl_domain` if you're on `na2` etc. in production.

Thoughts @j-mutter & @iloveitaly ?

jwebbstevens:
@tatecarr The hard cutover is essentially tied to when a user refreshes Sandbox. NetSuite started process in January of 2018. Such a refresh forces the update to use the new domain centered access instead of the old webservices.sandbox model.

It would be best to set the current standard since that's the standard that NetSuite is already on. If a user is using an old version say pre 2017.2 its really on them to maintain.  Trying to be data center agnostic is also the way to go if at all possible. It's a chicken/egg problem but since users in my experience don't keep up with the latest in releases in general its good to kind of force them to do so even if it breaks things because the user is using an outdate version.

I like the newer version because its essentially pushing users towards data center agnosticism and essentially forcing the adoption of TBA which is better for everyone.

README.md
```diff
@@ -69,7 +69,7 @@ NetSuite.configure do
   api_version	'2012_1'
 
   # optionally specify full wsdl URL (to switch to sandbox, for example)
-  wsdl          "https://webservices.sandbox.netsuite.com/wsdl/v#{api_version}_0/netsuite.wsdl"
+  wsdl          "https://system.netsuite.com/wsdl/v#{api_version}_0/netsuite.wsdl"
```
jrusso1020: should this be `"https://webservices.netsuite.com/wsdl/v#{api_version}_0/netsuite.wsdl"`? In order to get this working correctly for our current setup we had to use the former url not the `system` one

jwebbstevens: Correct.  You may have to add in the Data Center (ie. na1, na2) if users aren't going to use data center agnostic rules. 

From NetSuite Help Docs:

The format of an account-specific domain name is: <account ID>.<service>.netsuite.com. For example, if your account ID is 123456, your account-specific domains would be:
     123456.suitetalk.api.netsuite.com
     123456.restlets.api.netsuite.com
     123456.shop.netsuite.com

As of 2017.2, account-specific domains are supported for SuiteTalk (web services), RESTlets, and External Catalog (WSDK) shopping sites. These account-specific domains are unique to your account because they contain your account ID as part of the domain name. These domains do not change when your account is moved to a different data center.

Format of a Data Center-Specific NetSuite URL

For web services, rest, and External Catalog (WSDK) site URLs, consider using account-specific URLs rather than data center-specific URLs. See URLs for Account-Specific Domains.

Data center-specific NetSuite URLs are strings made up of the following parameters: <ServiceType>.<LocationId>.<AccountType>.netsuite.com.

cooncesean: Ignorant question: How do you find your account id?

iloveitaly: @cooncesean Navigate to the SuiteTalk preferences via “Setup > Integration > Web Services Preferences” to find this ID.

iloveitaly:
The best way to determine the correct wsdl url / endpoint for a given account ID (sandbox or production) is to use `NetSuite::Utilities.data_center_url('account_id')`. The readme has been updated to include more details about setting this up properly.

PR #388, opened 2018-02-26
Fix Samoa's NetSuite country value

hdeshev:
Use just `_samoa` instead of `_westernSamoa`

Very similar to the problem that #379 solved for Libya.

Linking the [docs](https://system.netsuite.com/help/helpcenter/en_US/srbrowser/Browser2017_2/schema/enum/country.html?mode=package) for the enum values for reference.


iloveitaly:
@hdeshev great catch, merging!

PR #394, opened 2018-03-28
Fix relationship between Customer and CustomerSalesTeamList

jeperkins4:
This story completes the proper relationship between Netsuite::Records::Customer and CustomerSalesTeamList.


iloveitaly:
@jeperkins4 thanks for this! Finally got around to merging it in. I made a couple changes, but it's merged into master with c9ac960...6659d54

PR #396, opened 2018-04-26
Remove duplicated Serbian country code

dlachasse:
This might be a little overwhelming to review. I removed this duplicated country code causing a constant override warning:

> `lib/netsuite/support/country.rb:187: warning: key "RS" is duplicated and overwritten on line 198`


iloveitaly:
@dlachasse haha! Totally overwhelming :)

PR #400, opened 2018-06-28
Use time zone support to avoid offset hacking

tatecarr:
I think at some point, Circle may have changed the Time Zone environment variable to UTC (or some zone that doesn't observe daylight savings).

After debugging, I think what was happening was that the utility function [line#270](https://github.com/NetSweet/netsuite/blob/master/lib/netsuite/utilities.rb#L270) was always returning `false` for `.dst?` so the offset was never changed from `-08:00` to `-07:00` (hence the failing test).

To test this theory, if you use a date that is within daylight savings, but is the UTC zone, it returns `false` for `.dst?`.  That same time, but in Pacific Time zone/offset, returns `true`.

And the main catch here is that `.to_time` returns the time in the local time zone, so even though the `.dst?` check in the utility function was after changing the time's offset to Pacific, it always evaluated to `false` because `.to_time` was always changing it back from PST to whatever the local zone is specified for the environment (I imagine UTC), and therefore didn't make the additional offset change (from `-08:00` to `-07:00) to be correct when the provided date was within daylight savings.

Instead of the current approach, manually changing the offset, then adding back the number of hours for the offset, I added active support which adds some time zone helpers, and now we can take the date associated with the provided `unix_timestamp`, use `Time.zone.parse(date)`, and the returned `Time` object will already have the correct offset set based on the date that was parsed.  Not sure how heavy a gem `activesupport` is, but this resolves the time issue, which probably wouldn't have manifested itself in production given how NS handles dates, but it works correctly now and all tests pass again.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,3 +9,5 @@ gem 'pry-rescue'
 # required for CircleCI to build properly with ruby 1.9.3
 gem 'json', '~> 1.8.3'
 gem 'rack', '~> 1.6.4'
+
+gem 'activesupport', '~> 5.0'
--- a/Gemfile
+++ b/Gemfile
@@ -10,4 +10,4 @@ gem 'pry-rescue'
 gem 'json', '~> 1.8.3'
 gem 'rack', '~> 1.6.4'
 
-gem 'activesupport', '~> 5.0'
+gem 'activesupport', '4.2.10'
--- a/lib/netsuite.rb
+++ b/lib/netsuite.rb
@@ -1,4 +1,5 @@
 require 'set'
+require 'active_support/all'
 
 require 'savon'
 require 'netsuite/version'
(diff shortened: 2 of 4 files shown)


tatecarr:
@iloveitaly Could fix the Failed circle status on the readme if this PR looks good.
image

iloveitaly:
@tatecarr sorry for the delay in getting back to you here! I have some updates locally to this that should fix the issue. The trick here is I don't want to include `activesupport` as a dependency—it brings with it a lot of packages which could cause dependency conflicts for users. The less external gems we use for this gem, the better.

Too busy to take a deep look here, but I'll get back to you here when I can!

iloveitaly:
Fixed this one up today with e880cbda46434f31f89cc8f176aa98a264945ad7!

PR #403, opened 2018-07-26
Throw a helpful error when netsuite returns the wrong thing

mseverini:
A few days ago, a Netsuite server went down, and when we tried to fetch data the server returned an error html page. The gem then through an unable to parse JSON error. This blew up our Netsuite integration with errors. These errors are really difficult to handle in their current state because they are just ruby type errors. By throwing a more distinctive error when we get an invalid response, it will make it far easier to handle an invalid Netsuite response.

### DIFF
--- a/lib/netsuite/response.rb
+++ b/lib/netsuite/response.rb
@@ -7,6 +7,8 @@ module NetSuite
       @header   = attributes[:header]
       @body     = attributes[:body]
       @errors   = attributes[:errors] || []
+
+      validate_response
     end
 
     def success!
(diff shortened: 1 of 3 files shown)


aom:
@mseverini I was working with response object and noticed that `body` can also be an Array for List Actions. Does all tests really pass for this PR?

lib/netsuite/response.rb
```diff
@@ -16,5 +18,9 @@ def success!
     def success?
       @success
     end
+
+    def validate_response
+      raise NetSuite::InvalidResponseError unless @body.is_a?(Hash)
```
aom: How about:

```
@body.is_a?(Hash) || @body.is_a?(Array)
```

mseverini: Nice catch!

mseverini:
some of the legitimate errors netsuite returns are xml, we will have to do this more intelligently, closing for now.

PR #406, opened 2018-08-16
Adds AddList to Actions

aom:
NetSuite supports AddList operation so this'll add it into the gem.

You could do same with UpsertList with the exception that if you're using externalId's you will possibly mutate existing records. AddList will throw duplicate error instead.

I took the liberty to update NetSuite version numbers in fixtures related to this new spec.

### DIFF
--- a/lib/netsuite/records/account.rb
+++ b/lib/netsuite/records/account.rb
@@ -7,7 +7,7 @@ module NetSuite
       include Support::Actions
       include Namespaces::ListAcct
 
-      actions :get, :get_list, :add, :update, :delete, :search, :upsert
+      actions :get, :get_list, :add, :add_list, :update, :delete, :search, :upsert
 
       fields :acct_name, :acct_number, :acct_type, :cash_flow_rate, :cur_doc_num, :description, :eliminate, :exchange_rate,
         :general_rate, :include_children, :inventory, :is_inactive, :opening_balance, :revalue, :tran_date, :balance
--- a/lib/netsuite/records/accounting_period.rb
+++ b/lib/netsuite/records/accounting_period.rb
@@ -6,7 +6,7 @@ module NetSuite
       include Support::Actions
       include Namespaces::ListAcct
 
-      actions :get, :get_list, :add, :delete, :upsert, :search
+      actions :get, :get_list, :add, :add_list, :delete, :upsert, :search
 
       fields :allow_non_gl_changes, :end_date, :is_adjust, :is_quarter, :is_year, :period_name, :start_date
 
--- a/lib/netsuite/records/assembly_build.rb
+++ b/lib/netsuite/records/assembly_build.rb
@@ -8,7 +8,7 @@ module NetSuite
 	  include Support::Fields
       include Namespaces::TranInvt
 
-      actions :get, :add, :initialize, :delete, :update, :upsert, :upsert_list,
+      actions :get, :add, :add_list, :initialize, :delete, :update, :upsert, :upsert_list,
         :search
 
       fields :bin_numbers, :buildable, :created_date, :expiration_date,
--- a/lib/netsuite/records/assembly_item.rb
+++ b/lib/netsuite/records/assembly_item.rb
@@ -7,7 +7,7 @@ module NetSuite
       include Support::Actions
       include Namespaces::ListAcct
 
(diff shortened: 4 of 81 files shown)


lib/netsuite/support/actions.rb
```diff
@@ -32,6 +32,8 @@ def action(name)
             self.send(:include, NetSuite::Actions::Search::Support)
           when :add
             self.send(:include, NetSuite::Actions::Add::Support)
+          when :add
+            self.send(:include, NetSuite::Actions::AddList::Support)
```
iloveitaly: @aom do we know that every record that supports `add` also supports `add_list`? My hunch would be this isn't the case, but I could be wrong here.

If every record does support `add_list` we should eliminate the double `when :add` and add `add` and `add_list` under the same `when`.

iloveitaly:
@aom Thanks for the contribution! Apologies for the delayed response. Left a comment on the PR, otherwise this looks awesome.

lib/netsuite/support/actions.rb
aom: This was not intentional. I'll change it to `when :add_list`.

AFAIK the addList is widely supported by all record types. At least I can't find any exceptions from NetSuite help section: https://system.eu2.netsuite.com/app/help/helpcenter.nl?fid=section_N3481360.html

lib/netsuite/actions/add_list.rb
```diff
@@ -0,0 +1,118 @@
...
+            response = NetSuite::Actions::AddList.call(netsuite_records, credentials)
+
+            if response.success?
+              response.body.map do |attr|
+                record = netsuite_records.find do |r|
+                  r.external_id == attr[:@external_id]
+                end
+
+                record.instance_variable_set('@internal_id', attr[:@internal_id])
+              end
```
mbianco-stripe: @aom what happens here if you don't set an externalID on the record you are running through `addList`? `record` would be `nil` and throw an exception, right?

Are the results returned by `addList` just a internal and external ID, or is the whole record returned?

Also feels weird to use `instance_variable_set` here, but without modifying every record in the gem there isn't a great alternative right now. The only thing I'd like to explore here is if we can use the results from `addList` to generate a new list of records to return. This would eliminate the `find` usage and `instance_variable_set`.

What do you think? I haven't worked with `addList` so I could be wrong here.

lib/netsuite/actions/add_list.rb
```diff
@@ -0,0 +1,118 @@
...
+
+            response = NetSuite::Actions::AddList.call(netsuite_records, credentials)
+
+            if response.success?
+              response.body.map do |attr|
+                record = netsuite_records.find do |r|
+                  r.external_id == attr[:@external_id]
+                end
+
+                record.instance_variable_set('@internal_id', attr[:@internal_id])
```
iloveitaly: @aom what happens here if you don't set an externalID on the record you are running through `addList`? `record` would be `nil` and throw an exception, right?

Are the results returned by `addList` just a internal and external ID, or is the whole record returned?

Also feels weird to use `instance_variable_set` here, but without modifying every record in the gem there isn't a great alternative right now. The only thing I'd like to explore here is if we can use the results from `addList` to generate a new list of records to return. This would eliminate the `find` usage and `instance_variable_set`.

What do you think? I haven't worked with `addList` so I could be wrong here.

lib/netsuite/support/actions.rb
iloveitaly: Ok, good to know! Awesome work on this PR. Excited to get it merged!

lib/netsuite/actions/add_list.rb
aom: @iloveitaly You might confusing AddList with UpsertList. You don't need to define externalId for AddList records because they are always added to NetSuite unless other validation errors occur. With UpsertList you'll need to define a unique externalId for new records or internalId for existing records.

Yes, unfortunately AddList returns just the reference to NetSuite Record with internalId and externalId. No luck there.

But you did point out a bug! 

I'll include another test case without externalId. Because `instance_variable_set` will throw an exception if it is called for `nil` record. As I said above, the externalId is not mandatory for AddList records (I use it anyway on my queries so I didn't fall into this trap).

aom: I added a safe guard in case add_list is being called without external_id.

aom:
@iloveitaly rebased this branch as well to resolve conflicts

PR #407, opened 2018-08-17
Fix bin number serialization

manmartinez:
#### What's this PR do?
With this change we'll make sure we're using the right namespace for bin numbers, according to the netsuite schema documentation, both `BinNumberList` and `BinNumber` are defined under the `listAcct` namespace.

We'll also extract the `BinNumber` class into its own file and define the proper fields for it.

#### Notes
My app is using `2017_2` as its API version I'm not 100% sure if these changes are applicable to previous versions, I used the following link for reference: http://www.netsuite.com/help/helpcenter/en_US/srbrowser/Browser2017_2/schema/other/inventoryitembinnumberlist.html?mode=package


iloveitaly:
@manmartinez looks great, thanks!

PR #414, opened 2018-10-18
Add missing countries

freezepl:
- Aland Islands
- Caribbean Netherlands: Bonaire, Sint Eustatius and Saba
- Canary Islands
- Ceuta and Melilla
- Saint Martin
- South Sudan

Additionally added Kosovo.
Kosovo is not listed as an ISO standard country.
The unofficial 2-digit codes are used by the European Commission and others until Kosovo is assigned an ISO code.

[NetSuite Country list](http://www.netsuite.com/help/helpcenter/en_US/srbrowser/Browser2016_1/schema/enum/country.html?mode=package) used to check missing codes.


iloveitaly:
@freezepl thanks for the contribution!

PR #421, opened 2018-12-20
Attempting to fix an issue when parsing custom fields in search results

LNRexpress:
I was getting something like this in my search results:
```
            <setupCustom:customFieldList>
              <platformCore:customField internalId="1171" scriptId="custrecord_188_cseg_division" xsi:type="platformCore:SelectCustomFieldRef">
                <platformCore:value internalId="1" typeId="189">
                  <platformCore:name>Division Name</platformCore:name>
                </platformCore:value>
              </platformCore:customField>
              <platformCore:customField internalId="1168" scriptId="custrecord_cseg_projectid_n117" xsi:type="platformCore:MultiSelectCustomFieldRef">
                <platformCore:value internalId="3" typeId="-117">
                  <platformCore:name>Subsidiary Name</platformCore:name>
                </platformCore:value>
              </platformCore:customField>
            </setupCustom:customFieldList>
```
The second custom field (with type `platformCore:MultiSelectCustomFieldRef`) was being processed as an `Array`, but it was parsed into a `Hash`. This caused the field value to be inaccessible.

I'm not sure if this is the best way to handle this situation, but figured I'd present this solution and see where it leads.


LNRexpress:
Created the following issue to better explain the situation:
https://github.com/NetSweet/netsuite/issues/424

iloveitaly:
Thanks for the contribution! This is fixed here:

https://github.com/NetSweet/netsuite/commit/3377c971d0cb727d81f4b4bc6e30edfbdfaccfd1

PR #429, opened 2019-05-20
Add direct_revenue_posting field to non-inventory sale item model

antoinesaliba:
Direct revenue posting is a boolean for non-inventory sale item that you can find here http://www.netsuite.com/help/helpcenter/en_US/srbrowser/Browser2018_2/schema/record/noninventorysaleitem.html


iloveitaly:
Thanks @antoinesaliba!!

PR #431, opened 2019-06-10
Added support case types and messages

jeremycrosbie:


iloveitaly:
@jeremycrosbie Great, thank you!

PR #435, opened 2019-09-04
rescue Exception => rescue StandardError

shaicoleman:
https://thoughtbot.com/blog/rescue-standarderror-not-exception


iloveitaly:
@shaicoleman nice change! Thanks.

PR #439, opened 2019-09-20
fix "must use account-specific domains" error with newer API versions

### LINKED ISSUE #430: Connection errors with the 2019_1 API
I've configured Netsuite with a wsdl_domain:
```
        api_version   '2019_1'
        wsdl_domain   '1234567.suitetalk.api.netsuite.com'
```

However, it seems to be connecting to `webservices.netsuite.com:80` instead.
```
DEBUG -- : HTTPI /peer GET request to 1234567.suitetalk.api.netsuite.com (net_http)
 INFO -- : SOAP request: http://webservices.netsuite.com/services/NetSuitePort_2019_1
 INFO -- : SOAPAction: "get", Content-Type: text/xml;charset=UTF-8, Content-Length: 1129
DEBUG -- : <?xml version="1.0" encoding="UTF-8"?>
<env:Envelope xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:platformMsgs="urn:messages_2019_1.platform.webservices.netsuite.com" xmlns:env="http://schemas.xmlsoap.org/soap/envelope/" xmlns:platformCore="urn:core_2019_1.platform.webservices.netsuite.com">
  <env:Header>
    <platformMsgs:tokenPassport>
      <platformCore:account>***FILTERED***</platformCore:account>
      <platformCore:consumerKey>***FILTERED***</platformCore:consumerKey>
      <platformCore:token>***FILTERED***</platformCore:token>
      <platformCore:nonce>***FILTERED***</platformCore:nonce>
      <platformCore:timestamp>1559923804</platformCore:timestamp>
      <platformCore:signature algorithm="HMAC-SHA256">***FILTERED***
</platformCore:signature>
    </platformMsgs:tokenPassport>
  </env:Header>
  <env:Body>
    <platformMsgs:get>
      <platformMsgs:baseRef xsi:type="platformCore:RecordRef" internalId="123456" type="file"/>
    </platformMsgs:get>
  </env:Body>
</env:Envelope>

DEBUG -- : HTTPI /peer POST request to webservices.netsuite.com (net_http)
Netsuite::Client::Error (Failed to open TCP connection to webservices.netsuite.com:80 (No route to host - connect(2) for "webservices.netsuite.com" port 80))
```

### LINKED ISSUE #434: Regardless of Config, All Calls Go to Wrong DataCenter
I am attempting to update our NetSuite gem's configuration so that our requests go to the "right" data center. 

However, regardless of what I enter into the initializer, all requests go to the wrong data center, using the 'generic' wsdl: `https://webservices.netsuite.com/services/NetSuitePort_2017_2` as it appears to drop the `na2.` from the 'correct' URL (see below). 

In my configuration/initializer, I have specified one of two options: 

```
# config/initializers/netsuite.rb
NETSUITE_CONFIG = YAML.load_file(File.join(Rails.root, 'config', 'netsuite.yml'))[Rails.env].symbolize_keys

NetSuite.configure do
  reset!

  api_version   '2017_2'

  wsdl          "https://webservices.na2.netsuite.com/services/NetSuitePort_#{api_version}"
  # I have tried, with the same results, using one of the options below instead
  # wsdl          "https://OURACCT#.suitetalk.api.netsuite.com/wsdl/v#{api_version}_0/netsuite.wsdl"
  # wsdl_domain   'OURACCT#.suitetalk.api.netsuite.com'

  sandbox false
  account           NETSUITE_CONFIG[:api_account]
  consumer_key      NETSUITE_CONFIG[:consumer_key]
  consumer_secret   NETSUITE_CONFIG[:consumer_secret]
  token_id          NETSUITE_CONFIG[:token_id]
  token_secret      NETSUITE_CONFIG[:token_secret]

  read_timeout 100_000
  # let's save this stuff to its own log
  log File.join(Rails.root, 'log/netsuite.log')

  soap_header 'platformMsgs:ApplicationInfo' => {
    'platformMsgs:ignoreReadOnlyFields' => true
  }
end
```

The long and short is that no matter what I have in the initializer/config file, the library *only* makes requests to "https://webservices.netsuite.com/services/NetSuitePort_2017_2". I have even tried as `wsdl "http://dumburls.com/i/hate/netsuite"` as a WSDL URL and it still just goes to the URL noted above. 

One final note: if I change the API Version in the config file, it DOES change the request to `https://webservices.netsuite.com/services/NetSuitePort_2015_1` but still ignores whatever I have for `wsdl` or `wsdl_domain`. 

Any direction or suggestions would be most appreciated and thank you!

q3aiml:
Newer API versions require using the account-specific domains such as
`<account-id>.suitetalk.api.netsuite.com`. The WSDLs for these versions,
however, specify the generic `webservices.netsuite.com` domain. When
`wsdl_domain` is provided use this to override the endpoint so the
account-specific domain can be used and users can successfully use
2019_1.

I create the client first without the `endpoint` override to allow
reading the default endpoint from the client. Then `wsdl_domain` is
substituted into the default which avoids hardcoding the path in the gem
(`/services/NetsuitePort_#{api_version}`).

If this seems too messy I can instead (in addition?) open a PR that adds
a `wsdl_override` option.

Thanks for looking.

Fixes #430 
Fixes #434

### DIFF
--- a/lib/netsuite/configuration.rb
+++ b/lib/netsuite/configuration.rb
@@ -27,6 +27,7 @@ module NetSuite
         log_level: log_level,
         log: !silent, # turn off logging entirely if configured
       }.update(params))
+      client.wsdl.endpoint = client.wsdl.endpoint.to_s.sub('//webservices.netsuite.com/', "//#{wsdl_domain}/")
       cache_wsdl(client)
       return client
     end
--- a/spec/netsuite/configuration_spec.rb
+++ b/spec/netsuite/configuration_spec.rb
@@ -94,6 +94,24 @@ describe NetSuite::Configuration do
       end
     end
 
+    context 'when the API and wsdl domain have been set on newer api versions' do
+      before do
+        config.sandbox = false
+        config.api_version '2019_1'
+        config.wsdl_domain = '4810331.suitetalk.api.netsuite.com'
+      end
+
+      it 'should correctly modify the full wsdl path' do
+        expect(config.wsdl).to eql('https://4810331.suitetalk.api.netsuite.com/wsdl/v2019_1_0/netsuite.wsdl')
+      end
+
+      it 'should override endpoint using wsdl_domain' do
+        # Override endpoint as netsuite wsdls starting with 2019_1 default to the webservices.netsuite.com
+        # endpoint even though it is not supported by newer api versions
+        expect(config.connection.wsdl.endpoint).to eql('https://4810331.suitetalk.api.netsuite.com/services/NetSuitePort_2019_1')
+      end
+    end
+
     context '#cache_wsdl' do
       it 'stores the client' do
         expect(config.wsdl_cache).to be_empty


q3aiml:
Note that you can also patch something like this into a codebase using the current version of the gem with something like this:

```ruby
module NetSuite
  module Configuration
    extend self

    alias :old_connection :connection
  
      client = old_connection(params, credentials)
      client.wsdl.endpoint = client.wsdl.endpoint.to_s.sub('//webservices.netsuite.com/', "//#{wsdl_domain}/")
      client
    end
  end
end
```

jormon:
Helped me out -- thank for this!

mawaldne:
Thank you! This also helped me.

iloveitaly:
Hey All! Still haven't had time to dig into this contribution and get it mixed it. Will hopefully get to this sooner rather than later.

ohaddahan:
@q3aiml 

```ruby
module NetSuite
  module Configuration
    extend self

    alias :old_connection :connection

      client = old_connection(params, credentials)
      client.wsdl.endpoint = client.wsdl.endpoint.to_s.sub('//webservices.netsuite.com/', "//#{wsdl_domain}/")
      client
    end
  end
end

NetSuite.configure do
  reset!
  account          ENV['NETSUITE_ACCOUNT_ID']
  consumer_key     ENV['NETSUITE_CLIENT_ID']
  consumer_secret  ENV['NETSUITE_CLIENT_SECRET']
  token_id         ENV['NETSUITE_TOKEN_ID']
  token_secret     ENV['NETSUITE_TOKEN_SECRET']
  api_version      ::NETSUITE_API_VERSION
  wsdl_domain      ::NETSUITE_WSDL_DOMAIN
end
```

In my `Rails` `config/initializers`, getting error:
> NoMethodError: undefined method `wsdl' for #<Savon::Client:0x00007fe4b7254010>

q3aiml:
@ohaddahan I haven't been using netsuite for some time so I am afraid I won't be much help. I'll close to this for clarity. If anyone who still has the ability to test this change wants to open a new PR please feel free to use my patch (as hacky as it is!).

darrend:
@ohaddahan (and others) 

(copying my comment verbatim from https://github.com/NetSweet/netsuite/pull/445#issuecomment-779965070)

To get going with `2020_2` (2020.2, are these twice a year like clockwork? do we need to keep up with them?) I had to refine the hacks above a bit. The urls mentioned in the wsdl returned by Netsuite changed again from `webservices.na0.netsuite.com` to `webservices.netsuite.com` so I tweaked the code in https://github.com/NetSweet/netsuite/pull/445#issuecomment-594861923 to use a more generic regex replacement. In addition, Savon doesn't expose (or no longer in the version included by this gem) `wsdl` so I had to crack it open and expose the attribute; this fixes the `undefined method `wsdl'` error.

Below is for a totally made up `0000000_SB1` sandbox account. Replace it with your real account of course.

```ruby
NetSuite.configure do
  reset!
  api_version      '2020_2'
  account          "0000000_SB1" 
  consumer_key     "xxxxx"
  consumer_secret  "xxxxx"
  token_id         "xxxxx"
  token_secret     "xxxxx"
  # pull the following from `NetSuite::Utilities.data_center_url('0000000_SB1')` every time? once year? never? always?
  wsdl_domain     "0000000-sb1.suitetalk.api.netsuite.com"
end

module NetSuite
  module Configuration
    extend self

    alias :old_connection :connection

      client = old_connection(params, credentials)
      wsdl = client.wsdl_netsuite_endpoint_hack
      # from "https://webservices.netsuite.com/services/NetSuitePort_2020_2" to "https://#{wsdl_domain}/services/NetSuitePort_2020_2"
      wsdl.endpoint = wsdl.endpoint.to_s.gsub(%r[//[^/]*/], "//#{wsdl_domain}/") #
      client
    end
  end
end

# above hack fails with undefined method `wsdl' for #<Savon::Client:0x00007fa07cb5c2e0> otherwise
module Savon
  class Client
    def wsdl_netsuite_endpoint_hack
      return self.wsdl if self.respond_to?(:wsdl)
      @wsdl
    end
  end
end
```

iloveitaly:
For any googlers, this was fixed with https://github.com/NetSweet/netsuite/pull/473

PR #443, opened 2019-11-20
Add the status field to ItemFulfillments

jormon:
Item Fulfillments in NetSuite have a `status` field that should be available via the record.


iloveitaly:
thanks!

Issue #447, opened 2020-02-28
Inconsistency with Segments


toyhammered:
Both [Department](https://github.com/NetSweet/netsuite/blob/05b8a16a60a26a35e4124bf9b315b9fef046c2fa/lib/netsuite/records/department.rb#L14) and [Location](https://github.com/NetSweet/netsuite/blob/05b8a16a60a26a35e4124bf9b315b9fef046c2fa/lib/netsuite/records/location.rb#L19) have a `record_refs: :parent` while [Classifcation](https://github.com/NetSweet/netsuite/blob/05b8a16a60a26a35e4124bf9b315b9fef046c2fa/lib/netsuite/records/classification.rb#L12) has a `fields: :parent`. 
Would it be possible to add it as a record_refs so it can be consistent with other segments?

iloveitaly:
@toyhammered yes! Want to submit a PR with this change?

PR #448, opened 2020-03-04
update classification segment to include parent record_ref

toyhammered:
This was done to keep Classification consistent with Location and Department segments.

Fixes https://github.com/NetSweet/netsuite/issues/447


iloveitaly:
@toyhammered Thanks! I made a quick change in addition to your PR, the `parent` in the `fields` array should be removed. 

Here's the SHA with that change: 1292ddce014df1ec2311d87fce49461c4db4b43f

toyhammered:
Perfect, thank you! Wasn't sure if the fields should be kept.

PR #454, opened 2020-04-23
Add amount as a read_only_field to Invoice

jormon:
The Invoice object has an amount property in netsuite that can be
returned in transaction searches.  It should be available here too.


iloveitaly:
Awesome, thank you! @jormon

PR #455, opened 2020-04-25
Add item vendor list to non inventory resale item

nkulbiski:
This adds the ability to add a Item Vendor List to a Non Inventory Resale Item.

It's a simple one line change.

I also added tests for Non Inventory Resale Item


iloveitaly:
@nkulbiski awesome, thank you!

PR #458, opened 2020-06-17
Pick up a commit from someone else's fork

jonkraut:
This commit has been sitting in a PR against NetSweet/NetSuite for a
few months, but is necessary for this Gem to be usable at all with the
version of the API that we will be using.

- But that branch/PR has fallen behind master
- And I'd rather point to OdekoTeam's github to pick up an unreleased
  version of a Gem, rather than someone else's repo.

When these changes are released into the upstream NetSweet/NetSuite
proper, we can point our Gemfile at that.

original commit message from:
https://github.com/q3aiml/netsuite/tree/override-endpoint-using-wsdl-domain
https://github.com/q3aiml/netsuite/commit/1b6093a223e5db2c4ab9bffc284e550d9ebba214

fix "must use account-specific domains" error with newer API versions

Newer API versions require using the account-specific domains such as
`<account-id>.suitetalk.api.netsuite.com`. The WSDLs for these versions,
however, specify the generic `webservices.netsuite.com` domain. When
`wsdl_domain` is provided use this to override the endpoint so the
account-specific domain can be used and users can successfully use
2019_1.

I create the client first without the `endpoint` override to allow
reading the default endpoint from the client. Then `wsdl_domain` is
substituted into the default which avoids hardcoding the path in the gem
(`/services/NetsuitePort_#{api_version}`).

### DIFF
--- a/lib/netsuite/configuration.rb
+++ b/lib/netsuite/configuration.rb
@@ -27,6 +27,7 @@ module NetSuite
         log_level: log_level,
         log: !silent, # turn off logging entirely if configured
       }.update(params))
+      client.wsdl.endpoint = client.wsdl.endpoint.to_s.sub('//webservices.netsuite.com/', "//#{wsdl_domain}/")
       cache_wsdl(client)
       return client
     end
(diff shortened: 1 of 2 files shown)


jonkraut:
Apologies. I had the wrong screen open when I made the PR

PR #462, opened 2020-08-04
Allow `NetSuite::Configuration#log_level` to be reset

dleve123:
## Problem

It’s surprising to not be able to, in a rails console, increase
the verbosity of the gem’s logging without needing to:

1. Tweak an ENV variable or otherwise change an initializer
2. Directly set `Configuration#attributes`

Aka, I would expect the following to work, but it does not.

```
app:staging> NetSuite::Configuration.log_level
=> :info
app:staging> NetSuite::Configuration.log_level = :debug
=> :debug
app:staging> NetSuite::Configuration.log_level
=> :info
```

Seems like the only way to increase the verbosity is to directly set
the appropriate attribute, which seems to be an internal data structure not intended
to be used by client code.

```
gravity:staging> NetSuite::Configuration.attributes[:log_level]
=> :info
gravity:staging> NetSuite::Configuration.attributes[:log_level] = :debug
=> :debug
gravity:staging> NetSuite::Configuration.log_level
=> :debug
```
## Solution

* Backfills some test coverage (f6fa522)
* Always set `attributes[:log_level]` when `log_level=` is called
* Support `log_level(value)` syntax via a short-circuit, similar to
  what's done in `#log(path)` and `#silent(value)`

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,13 @@ NetSuite.configure do
   read_timeout  100_000
 
   # you can specify a file or file descriptor to send the log output to (defaults to STDOUT)
+  # If using within a Rails app, consider setting to `Rails.logger` to leverage existing
+  # application-level log configuration
   log           File.join(Rails.root, 'log/netsuite.log')
+
+  # Defaults to :debug level logging for Savon API calls. Decrease the verbosity
+  # by setting log_level to `:info`, for example
+  # log_level   :debug
 
   # password-based login information
   email    	  'email@domain.com'
--- a/lib/netsuite/configuration.rb
+++ b/lib/netsuite/configuration.rb
@@ -373,12 +373,13 @@ module NetSuite
     end
 
     def log_level(value = nil)
-      self.log_level = value || :debug
-      attributes[:log_level]
+      self.log_level = value if value
+
+      attributes[:log_level] || :debug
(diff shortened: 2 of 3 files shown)


dleve123:
Seems like the tests are broken on `master`, so here's some evidence that the configuration object still works!

```
dl@phoenix:~/src/netsuite (allow-log-level-to-be-reset)
$ rspec spec/netsuite/configuration_spec.rb

Randomized with seed 60579
..................**..........................

Pending: (Failures listed here are expected and do not affect your suite's status)

  1) NetSuite::Configuration#credentials when none are defined should properly create the auth credentials
     # No reason given
     # ./spec/netsuite/configuration_spec.rb:329

  2) NetSuite::Configuration#credentials when they are defined should properly replace the default auth credentials
     # No reason given
     # ./spec/netsuite/configuration_spec.rb:333


Finished in 0.03101 seconds (files took 0.67646 seconds to load)
46 examples, 0 failures, 2 pending

Randomized with seed 60579
```

---

Hope this is helpful / desirable. If not, no worries. Thanks for dealing with SOAP APIs for us ;)

cc @iloveitaly

iloveitaly:
Great change! Thanks for the PR.

PR #463, opened 2020-08-14
Add fields to Partner and Employee records

nkulbiski:
Add `subsidiary` to partner record
Add `sales_role` to employee record

Update tests for employee record
Add tests for partner record


iloveitaly:
Great change! Thanks.

PR #466, opened 2020-12-15
CustomerPaymentCreditList

stevenou:
Kinda confused because I didn't see this `Record` type, however, the following already existed in `customer_payment_spec.rb`:

```
  describe '#credit_list' do
    it 'can be set from attributes'
    it 'can be set from a CustomerPaymentCreditList object'
  end
```


iloveitaly:
Great change! Thank you. There are some other follow-on changes that I'll make real quickly after merging and post here in case you have future similar PRs.

iloveitaly:
Here are the couple of other changes that needed to be made 776dc1b...dfab101

stevenou:
Awesome, thanks for the quick merge!

PR #468, opened 2021-02-11
Update fields/record refs on partner record to match 2020.2 schema

cgunther:
Thanks for this gem. I started my first integration with NetSuite and this was a huge time saver!

My integration touched on partners and I noticed the field list and record refs were outdated compared to the 2020.2 schema, so this should bring the gem inline with that schema.


iloveitaly:
Another great fix! Thank you.

cgunther:
Thanks for the fast merge!

PR #470, opened 2021-02-14
Support non-record field results in saved search responses - Revised

cgunther:
This is a continuation of #426, addressing some snags I had using it to address a similar problem (https://github.com/NetSweet/netsuite/pull/426#issuecomment-778261253). Thanks @davidlaprade for the great starting point.

`internal_id` and `external_id` are now extracted from the search result to the record.

For API versions 2013.2 and newer, the non-standard fields returned in search are added as custom fields on the record, using the field name as `script_id`.

In wrapping this up, I'm noticing that both the real custom fields and the non-standard search result fields wrap their value in a `searchValue` element, however typically for custom fields, it seems to be wrapped in a `value` element, which the record then defines a helper method to access:
https://github.com/NetSweet/netsuite/blob/438c233fc0ede0acbeb7f9fbe8c9d8004fb4654a/lib/netsuite/records/custom_field.rb#L17-L19

As this PR stands now, you'd access a non-standard search result field (or custom field) via: `my_record.custom_field_list.my_custom_field.attributes.fetch(:search_value)`. However, after a typical `get` action, you'd access the same custom field via `my_record.custom_field_list.my_custom_field.value`. This doesn't appear to be a regression due to this PR, just an existing issue that may be surfacing now. When dealing with search results, would it make sense to copy the `:search_value` attribute to `:value` to ease using the same `#value` helper?

I'm still new to NetSuite so I'm not sure if I'm overlooking something here.

### DIFF
--- a/lib/netsuite/support/search_result.rb
+++ b/lib/netsuite/support/search_result.rb
@@ -98,7 +98,7 @@ module NetSuite
                       # otherwise it will be lost when we initialize
                       custom_fields = record[search_group][:custom_field_list][:custom_field]
                       custom_fields = [custom_fields] if custom_fields.is_a?(Hash)
-                      custom_fields << search_result.merge(internal_id: attr_name)
+                      custom_fields << search_result.merge(NetSuite::Records::CustomFieldList.reference_id_type => attr_name)
                       record[search_group][:custom_field_list][:custom_field] = custom_fields
                     end
                   else
--- a/lib/netsuite/support/search_result.rb
+++ b/lib/netsuite/support/search_result.rb
@@ -59,6 +59,9 @@ module NetSuite
                 # skip all attributes: look for :basic and all :xxx_join
                 next if search_group.to_s.start_with?('@')
 
+                # avoids `RuntimeError: can't add a new key into hash during iteration`
+                record[search_group][:custom_field_list] ||= {custom_field: []}
+
                 record[search_group].each_pair do |attr_name, search_result|
                   # example pair:
                   # {
@@ -93,7 +96,6 @@ module NetSuite
                     else
                       # not a record field -- treat it as if it were a custom field
                       # otherwise it will be lost when we initialize
-                      record[search_group][:custom_field_list] ||= {custom_field: []}
                       custom_fields = record[search_group][:custom_field_list][:custom_field]
                       custom_fields = [custom_fields] if custom_fields.is_a?(Hash)
                       custom_fields << search_result.merge(internal_id: attr_name)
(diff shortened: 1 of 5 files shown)


lib/netsuite/support/search_result.rb
```diff
@@ -59,11 +59,22 @@ def initialize(response, result_class, credentials)
...
 
-                record[search_group].each_pair do |k, v|
+                # avoids `RuntimeError: can't add a new key into hash during iteration`
+                record[search_group][:custom_field_list] ||= {custom_field: []}
+
+                record[search_group].each_pair do |attr_name, search_result|
+                  # example pair:
+                  # {
+                  #   :department=>{
+                  #     :search_value=>{:@internal_id=>"113"},
```
iloveitaly, line 69: `search_value` is the value of the search result, right? Not the input into the origin search query?

iloveitaly:
@cgunther Thanks for the work here! This is definitely something I'd like to merge in. I'll need to test this against some larger projects with test suites that would give me more confidence in this change.

I'm not sure I like stuffing non-custom field values in the customFieldList. Curious your thoughts on adding those as `read_only_fields` or a new field type (`search_only_fields`). To me, that seems like a better approach albeit more work for us on the frontend. My sense is adding these to the custom field list may cause issues down the line and will be confusing for users ("why is a non-custom field being put in the custom field list?")/ 

> When dealing with search results, would it make sense to copy the `:search_value` attribute to `:value` to ease using the same `#value` helper?

Yes, I'd opt for this approach. `search_value` seems to be yet another weird NetSuite-ism that we should try to shield users from as much as we can.

cgunther:
I had similar concerns about hijacking custom fields, but figured I'd leverage @davidlaprade's work first.

While conceptually I'd agree these are read-only fields, I'm hesitant to hijack that concept as well as I feel there's a clear difference between the read-only fields on an actual record vs these search-only fields that are really more specific to the type of search you're doing (technically `closeDate` is a field of `TransactionSearchBasic`, not `Invoice`). For example, invoices have a slew of [read-only fields](https://github.com/NetSweet/netsuite/blob/33594f42f0e4f5155f2b426e516c16deda4c50fd/lib/netsuite/records/invoice.rb#L42), but you get those back in a simple `get` call, whereas something like `closeDate` you'd only get back via a `search`.

I'll take a stab as reworking this towards a new field type. Longer-term it may be wise to encapsulate these search-only fields in modules mirroring the NetSuite searches (ie. `TransactionSearchBasic`), since especially in this case, that search is applicable to many record types (see "Related Record(s)" at the bottom https://system.netsuite.com/help/helpcenter/en_US/srbrowser/Browser2020_2/schema/search/transactionsearchbasic.html?mode=package)

cgunther:
Superseded by #483.

PR #474, opened 2021-03-24
Accounting period closed

stevenou:
Only way to get the `closed` attribute (which does not appear on a simple `get`) for an `AccountingPeriod` is to request it as a column as part of an Advanced Search. This makes that possible.

### DIFF
--- a/lib/netsuite/records/accounting_period.rb
+++ b/lib/netsuite/records/accounting_period.rb
@@ -8,7 +8,7 @@ module NetSuite
 
       actions :get, :get_list, :add, :delete, :upsert, :search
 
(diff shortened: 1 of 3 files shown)


iloveitaly:
@stevenou somehow this PR has a merge conflict. If you could rebase on master, I'd love to merge this change in.

stevenou:
hmmm I'm not sure I'm doing this right. I don't see a conflict on GitHub. if you still do, let me just create a new branch. the changes are simple enough.

iloveitaly:
Ah, was looking at the wrong view. Change looks great, thanks!

PR #477, opened 2021-05-25
Include protocol in readme examples setting endpoint from wsdl_domain

cgunther:
The `wsdl_domain` is just a domain/hostname, no protocol, however `endpoint` expects to include a protocol, so when you're buliding the `endpoint` from the `wsdl_domain`, you need to prefix it with a protocol.

#473 introduced the ability to set the `endpoint`.


iloveitaly:
Thanks!

PR #478, opened 2021-05-30
Properly extract external_id from advanced search results

cgunther:
If you performed an advanced search specifying which columns to return and included externalId, you wouldn't be able to get the external_id from the resulting record via the usual external_id method. This now matches how internal_id is extracted.

Prior to this, calls to external_id returned:
```
> record.external_id
=> {:@external_id=>"customer_1"}
```

Now, it returns:
```
> record.external_id
=> "customer_1"
```

This originally surfaced in #470 as part of a larger change, but we're exploring taking that in a different direction, so extracted this out to it's own PR.


iloveitaly:
Awesome fix! Thanks.

PR #480, opened 2021-07-12
Fix accessing custom field values returned in advanced search results

cgunther:
NetSuite seems to wrap values in a `searchValue` element rather than just `value`, breaking the ability to access those values the same in an advanced search as compared to either a basic search or simple `get`.

This was already handled for standard fields, but not custom fields:
https://github.com/NetSweet/netsuite/blob/0d1f5ab0ced599abdc0767eaa74ee6e04fc2122b/lib/netsuite/support/search_result.rb#L81

Previously, you'd access the value via `my_record.custom_field_list.my_custom_field.attributes.fetch(:search_value)`, and if it was a `CustomRecordRef`, you'd get a raw hash back, not an instance of `CustomRecordRef`.

Now, you'd access the value via `my_record.custom_field_list.my_custom_field.value`, so it's consistent across `get`, basic, and advanced searches, plus you'd get an instance of `CustomRecordRef`, where appropriate.

This originally surfaced in #470, but we're taking that in a different direction, so figured I'd address this separately.


iloveitaly:
Great change! Thanks. Took a long time to review and get merged because I wanted to run some manual tests against some netsuite accounts. Sorry about the delay.

cgunther:
All good! Thanks for the merge.

PR #482, opened 2021-07-13
Add CostCategory record

cgunther:
This is an accounting list, like Term.

I copied the get, add, delete specs from the Term spec, but it seems a little redundant to repeat these on each record. Maybe it'd be better to have an "action matcher", like the "field matcher", to assert the right actions were defined, not testing the implementation of those actions, leaving that to the action-specific specs?


iloveitaly:
Great change, thanks!

PR #483, opened 2021-08-07
Introduce search only fields

cgunther:
When performing a search, there's additional columns that can be requested that aren't normally on the record. Previously such columns were not accessible via this gem's search results.

Now these columns can be accessed just as you'd access a regular field or read only field.

Starting with InventoryItem and Invoice.

Ultimately, these are fields of the ___SearchRowBasic (ie. ItemSearchRowBasic and TransactionSearchRowBasic), so eventually it may be wise to extract a module encapulating those SearchRowBasic's and include that in the relevant records to avoid duplication of search only fields across multiple records (ie. Estimate, SalesOrder, Invoice, CreditMemo, etc.) that all utilize the same SearchRowBasic.

This is a second attempt at #470, which itself was a second attempt at #426.


iloveitaly:
Sorry for the lack of responses on the other PRs. I _really_ like this approach. Super clean. Thanks!

iloveitaly:
@cgunther it would be great to add some readme documentation around this if you have time for another PR.

cgunther:
Thanks for the quick merge!

PR #484, opened 2021-08-11
Explain search only fields in Readme

cgunther:
Attempt to document #483.

I wasn't sure how best to convey this given it affects consuming the search results, not performing the search itself. It should also be unsurprising that you'd access these search-only fields the same way you'd access regular or read-only fields.

Had we buried these search-only fields in custom fields, like #426 first attempted, then I'd find that more surprising and warranting clearer documentation.


iloveitaly:
This is great! Just a simple one or two lines to push folks in the right direction is helpful.

PR #485, opened 2021-08-14
return-netsuite-error-for-search-action-failures

Timothyjb:
Why
Currently, if the response of the `Search` action is not successful, it will simply return `false` and give no insight into what the issue was.
Looking at patterns implemented elsewhere in the codebase, we seem to be moving towards returning a `NetSuite::Error` if the response is not successful.

Solution
Implement returning a `NetSuite::Error` if the response is not successful.

### DIFF
--- a/lib/netsuite/actions/search.rb
+++ b/lib/netsuite/actions/search.rb
@@ -244,7 +244,7 @@ module NetSuite
             if response.success?
               NetSuite::Support::SearchResult.new(response, self, credentials)
             else
-              false
+              NetSuite::Error.new(response.body[:status][:status_detail])
             end
           end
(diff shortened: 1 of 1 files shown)


iloveitaly:
@Timothyjb where else in the codebase are you seeing this pattern?

I think this, combined with something like #405, is the right approach going forward. Just want to think on/validate the approach a bit more when I have some time.

aom:
@iloveitaly Would be awesome to see some progress on this after such a long time. We've been running a fork of this repository to get this feature to our production.

iloveitaly:
@aom Me too! Just time constrained on my end :/

PR #488, opened 2021-08-17
Avoid repeating field definitions

### LINKED ISSUE #486: Search only fields are not added to an upsert/add payload
@cgunther it looks like some of the search-only fields that were added are also non-search only fields (`other_ref_num`) is one example.

What do you think about checking if the field already exists within the `search_only_field` logic and not adding it if it's already been set as another field type?

cgunther:
In adding the start of search_only_fields, a couple standard fields were mistakenly re-defined as search_only_fields, which ended up excluding. them when pushing data to NetSuite, under the guise they were search-only.

Now it's impossible to define the same field twice.

As a result, had to fix a number of occurrences of fields repeated between `fields` and another definition (ie `record_refs`, `read_only_fields`, `field`).

The only notable fix is on `Invoice`, where `bill_address` and `ship_address` were previously defined as `fields`, then got repeated as `search_only_fields`. In old APIs (<= 2014.1), these seemed to actually be fields, however in newer APIs (>= 2014.2), they no longer are. I'm erring on the side of newer APIs here, removing them as `fields` and leaving them as `search_only_fields`.

Fixed #486

### DIFF
--- a/lib/netsuite/records/assembly_unbuild.rb
+++ b/lib/netsuite/records/assembly_unbuild.rb
@@ -12,7 +12,7 @@ module NetSuite
         :search
 
       fields :bin_numbers, :built, :created_date, :expiration_date,
-        :last_modified_date, :memo, :quantity, :serial_numbers, :total,
+        :last_modified_date, :memo, :quantity, :serial_numbers,
         :tran_date, :tran_id
 
 	  read_only_fields :total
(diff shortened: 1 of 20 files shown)


iloveitaly:
@cgunther great change, thanks!

Re: invoice change, this shouldn't cause any breaking changes for users unless they are on a 2014 API version. I think biasing towards the most recent API version makes sense. Would be good to note this in the changelog though.

PR #490, opened 2021-10-03
Updates README.md to reflect task.refresh, not task.reload

TimJKStrickland:
Couldn't figure out why .reload wasn't working, but it's because the method is called refresh, not reload.


iloveitaly:
good catch! thanks.

PR #496, opened 2021-10-20
Update Estimate fields/record_refs for 2021.2

cgunther:

This effectively removes the `balance` and `bill_is_residential` fields.
Best I can tell going back to 2014.1, they were never fields. Perhaps
this was a bad copy-paste when first introducing `Estimate` using
another  record as a starting point?

`is_multi_ship_to` was also removed as it's technically a field of
`TransactionSearchRowBasic`, not `Estimate`. It never could have been
set, and at best it only could have been read after a search. Ideally
it'd be reintroduced in the future as an expansion on the work in #483,
extracting a common module to represent the fields from
`TransactionSearchRowBasic`.

`bill_address` was also removed as it last appeared in 2014.1. Now it's a
field of `TransactionSearchRowBasic`, so it could be re-introduced
later, like above.

`billing_schedule` was corrected to be a `record_ref`.

`accountingBookDetailList`, `partnersList`, `salesTeamList`,
`shipGroupList`, and `taxDetailsList` are still missing as `record_refs`
as their corresponding classes haven't been implemented yet.

I improved the `have_field` matcher to optionally take a class argument
for testing the fields that are represented by special classes.

I deviated from the standard style of multiple-fields-per-line-wrapped
for a single-field-per-line style. I found the old style hard to read,
particularly when the fields fell out of alphabetical order, when
scanning to see either what fields were available, or what fields were
already supported. I'd imagine this'll make for cleaner git diffs in the
future too.


iloveitaly:
One field per line is a great idea.

Thanks for the clear explanation about the motivation behind this change.

Keep these coming!

Issue #363, opened 2017-10-10
Account search breaks on savon > 2.11.1


iloveitaly:

iloveitaly:
fixed with https://github.com/NetSweet/netsuite/pull/497

PR #497, opened 2021-11-16
Updating necessary Savon version

andrewdicken-stripe:
Savon has released new versions that have fixed the bug that previously prevented us from upgrading beyond version 2.11.1. We now don't need that unnecessary restriction and can work with newer Savon versions including the latest of 2.12.1.


iloveitaly:
Thanks! There was an error on savon 2.11: https://github.com/NetSweet/netsuite/issues/363. I don't think there is a way to indicate that only savon 2.11.x cannot be used in the gemspec, so I'll leave this comment here as a warning to any gem users.

PR #498, opened 2021-11-29
added gem 'tzinfo-data' to pass all test for ruby 3.0.0

gbs4ever:


iloveitaly:
This shouldn't be a required gem for CI run, but I can fix that separately via a matrix variable. Thanks for the contribution!

PR #500, opened 2021-12-13
Update ServiceResaleItem fields/record_refs for 2021.2

cgunther:

The following "fields" were removed as "fields" since they require
specialized record classes that haven't been implemented yet:
- cost_estimate_type
- item_options_list
- matrix_type
- out_of_stock_behavior
- overall_quantity_pricing_type
- presentation_item_list
- site_category_list
- sitemap_priority
- translations_list
- vsoe_deferral
- vsoe_permit_discount
- vsoe_sop_group

I arbitrarily went back 5 years (to version 2016.1) and these were all
using specialized types back then, so this doesn't appear to have been a
recent change on the NetSuite side of things that might affect users of
this gem.


iloveitaly (approved):
As always, thanks for the great PR!

PR #506, opened 2021-12-17
Update InventoryItem fields/record_refs for 2021.2

cgunther:

`member_list` was removed as a field as it doesn't exist on
InventoryItem in the NetSuite schema. It was added back in a108ae4d to
appease searching, at a time when KitItem inherited from InventoryItem,
which is no longer the case. If I'm understanding the ItemSearchRowBasic
schema correctly, it's not returned as a field there:
https://system.netsuite.com/help/helpcenter/en_US/srbrowser/Browser2021_2/schema/search/itemsearchrowbasic.html?mode=package


iloveitaly:
> `member_list` was removed as a field as it doesn't exist on
> InventoryItem in the NetSuite schema.

It's included on 5 kit-like items but I spot checked a couple and it looks like it's been included on those. Merging this in! Thanks.

PR #245, opened 2016-02-25
Support Attaching Files

jordan-allan:
- Adds support for attaching files
- Action added to Invoice

PR #509, opened 2021-12-28
Support Attaching Files - Revised

cgunther:
This is a continuation of #245, addressing some snags I had using it to address a similar need (https://github.com/NetSweet/netsuite/pull/245#discussion_r732884628). Thanks @jordan-allan for the great starting point.

The conflict between `Records#type` and `Invoice.search_only_field :type` is now resolved and there's a check to ensure future conflicts between public/protected methods and fields aren't introduced. More on this is the commit message.

The casing of the record type is also resolved to be lower camelcase instead of just lowercase.

Broadly I have some concerns about calling this action `attach_file` as NetSuite actually calls it `attach` (with a corresponding `detach`) and can be used for both files and contacts as the thing being attached to another record. My immediate need was for attaching files and that aligned with @jordan-allan's work, so I stuck on that route. Thinking ahead, if we wanted to support attaching contacts, would we introduce a new action `attach_contact`, or would we rename `attach_file` to `attach`, leading to a breaking change/deprecation for anyone using `attach_file`? To support contacts, I believe the only thing that would have to change is where the `attachedRecord` `@type` is hardcoded to "file" to pull the type from the record itself (and rename the `@file` variable accordingly).


lib/netsuite/support/fields.rb
```diff
@@ -22,6 +22,8 @@ def fields(*args)
         def field(name, klass = nil)
           name_sym = name.to_sym
           raise "#{name} already defined on #{self.name}" if fields.include?(name_sym)
+          raise "#{name} conflicts with a method defined on #{self.name}" if method_defined?(name_sym)
```
iloveitaly, line 25: Nice addition

lib/netsuite/support/records.rb
```diff
@@ -49,7 +49,15 @@ def to_attributes!(hash, kname, v)
...
-        "#{record_namespace}:#{self.class.to_s.split('::').last}"
+        "#{record_namespace}:#{record_type_without_namespace}"
+      end
+
+      def netsuite_type
+        record_type_without_namespace.lower_camelcase
+      end
+
+      def record_type_without_namespace
+        "#{self.class.to_s.split('::').last}"
```
iloveitaly, line 60: Yes!! I've been wanting to get something like this in the codebase for some time!

iloveitaly (approved):
Thank you!

> Thinking ahead, if we wanted to support attaching contacts, would we introduce a new action attach_contact, or would we rename attach_file to attach, leading to a breaking change/deprecation for anyone using attach_file?

I'd be in favor of using `attach_contact`. `attach` is very vauge, and having more specific names for these methods feels more descriptive than a general attach, and would allow us to add type-specific logic more easily. I think this is worth the tradeoff of drifting from the official NetSuite name for this method.

iloveitaly:
@cgunther one thing I'd love to see in the future are some readme additions to describe how to use this functionality.

Issue #164, opened 2015-02-26
Issue with CustomRecord saved search


spurnaye:
Hello,
Here is the code I am using to pull a custom search.

results =  NetSuite::Records::CustomRecord.search({
  saved: <search_internal_id>,   # your saved search internalId
  basic: [{
    field: 'recType',
            operator: 'is',
            value: ::NetSuite::Records::CustomRecordRef.new(internal_id: <internal_id>)
    }]
}).results

I see the successful output from Netsuite as xml is correctly getting printed with success flag. But I see following error:
gems/ruby-2.1.3/gems/netsuite-0.3.3/lib/netsuite/support/search_result.rb:64:in `block in initialize': undefined method `search_joins=' for #<NetSuite::Records::CustomRecord:0x007fdddb1ab510> (NoMethodError)
	from .rvm/gems/ruby-2.1.3/gems/netsuite-0.3.3/lib/netsuite/support/search_result.rb:40:in `each'
	from .rvm/gems/ruby-2.1.3/gems/netsuite-0.3.3/lib/netsuite/support/search_result.rb:40:in `initialize'
	from .rvm/gems/ruby-2.1.3/gems/netsuite-0.3.3/lib/netsuite/actions/search.rb:241:in `new'
	from .rvm/gems/ruby-2.1.3/gems/netsuite-0.3.3/lib/netsuite/actions/search.rb:241:in `search'
	from test.rb:88:in `<main>'

feigner:
I had a similar issue doing a search against InventoryItems. I monkeypatched the record class by adding `attr_accessor :search_joins` and things seem to be working for me.

feigner:
Also, if you don't want to change that class, I believe that you can avoid a search join by specifying `returnSearchColumns: false` in your search preferences.

iloveitaly:
This is fixed for CustomRecords with https://github.com/NetSweet/netsuite/commit/7801251c21379ddb9332f205f8aee7c30f3f1b5a

Feel free to submit a PR for the InventoryItem fix! Also, would be great to have a PR with some documentation on how `returnSearchColumns`works.

PR #511, opened 2022-01-20
Move definition of `search_joins` attribute from records to search action

cgunther:
Previously it was up to each record to define this attribute, if it
offered the search action. This created scenarios where a record like [`Bin`](https://github.com/NetSweet/netsuite/blob/9f1000535d62045e1fcb6220192840c943eaba57/lib/netsuite/records/bin.rb)
offered the `search` action, but didn't define the `search_joins`
attribute and therefore searching bins would error:

```
undefined method `search_joins=' for #<NetSuite::Records::Bin:0x00007f8f32948bf8 @internal_id="1044", @external_id=nil, @attributes={:bin_number=>"WIP-Fabric"}> (NoMethodError)
```

due to `SearchResult` assuming the record had such an attribute:
https://github.com/NetSweet/netsuite/blob/9f1000535d62045e1fcb6220192840c943eaba57/lib/netsuite/support/search_result.rb#L112

Now adding the `search` action to a record automatically adds the
necessary `search_joins` attribute.

The attribute was previously defined on `AssemblyComponent`,
`SerializedInventoryItemLocation`, and `WorkOrderItem`, however none of
those support actions at all, let alone the `search` action. They're all
records for sublist items. I suspect the attribute was erroneously
introduced in the first place as a result of copy/pasting to create the
record.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -15,7 +15,7 @@
 
 ### Fixed
 * Fix "undefined method `[]` for #<Nori::StringIOFile>" when adding File (#495)
-*
+* Moved definition of `search_joins` attribute from records to search action. The attribute was removed for AssemblyComponent, SerializedInventoryItemLocation, and WorkOrderItem as they don't offer the search action. (#511)
 
 ## 0.8.10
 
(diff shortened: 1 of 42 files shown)


cgunther:
This has been added piecemeal, record-by-record, in the past (#164, #280, #295, #334, #342, #349, #350, #412), but hopefully this now solves it once-and-for-all across all records, now and into the future.

iloveitaly (approved):
Much cleaner! Nice change.

PR #516, opened 2022-02-08
Support translations records

fabiensebban:
TL;DR -- This supports translations for records that have access to

---

This allow to get translations from objects. Here is an example of translation that can be found in an item:
```xml
 <listAcct:translationsList>
      <listAcct:translation>
        <listAcct:locale>_englishUK</listAcct:locale>
        <listAcct:language>English (UK)</listAcct:language>
        <listAcct:displayName>Display Name</listAcct:displayName>
        <listAcct:salesDescription>Sales description</listAcct:salesDescription>
      </listAcct:translation>
 </listAcct:translationsList>
```

As each record can have its own translation fields, the `Translation` records fields are generated dynamically.

### DIFF
--- a/lib/netsuite/records/translation.rb
+++ b/lib/netsuite/records/translation.rb
@@ -1,0 +1,13 @@
+module NetSuite
+  module Records
+    class Translation
+      include Support::Fields
+      include Support::Records
+      include Namespaces::ListAcct
+
+      def initialize(attributes = {})
+        initialize_from_dynamic_attributes_hash(attributes)
+      end
+    end
+  end
+end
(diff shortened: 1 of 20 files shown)


lib/netsuite/records/translation.rb
```diff
@@ -0,0 +1,13 @@
+module NetSuite
+  module Records
+    class Translation
+      include Support::Fields
+      include Support::Records
+      include Namespaces::ListAcct
+
+      def initialize(attributes = {})
+        initialize_from_dynamic_attributes_hash(attributes)
```
iloveitaly, line 9: @fabiensebban  Why did this need to use a new dynamic attributes approach? Are the fields included here dynamic?

iloveitaly:
Thank you!

PR #524, opened 2022-03-22
retry http client error subclasses since we use include? instead of is_a? on these

ericcj:


iloveitaly:
Great change, thanks!

Issue #398, opened 2018-06-19
Support nullFieldList


iloveitaly:
For fields, including customFields, you cannot pass Null values or empty tags to the NetSuite API. The proper way of deleting a value from a certain field, is to add it to the nullFieldList (which is a separate element within the request profile). Note, you insert the name of the element you wish to blank out here. In the case of customFields, you should use the NetSuite internal ID, not the human readable name from the profile

https://system.na1.netsuite.com/app/help/helpcenter.nl?fid=section_N3527090https://system.na1.netsuite.com/help/helpcenter/en_US/Output/Help/section_N3438152.htmlhttps://system.na1.netsuite.com/app/help/helpcenter.nl?fid=section_N3438152

iloveitaly:
fixed with https://github.com/NetSweet/netsuite/pull/529

PR #529, opened 2022-04-06
Add support for NullFieldList (starting with credit memos and invoices)

cgunther:
NullFieldList is used to either clear an existing value on update, or
avoid a default being applied on adding a record.

My use case only required support for credit memos and invoices, so I
started there, but this could likely be added to most records.

This was heavily inspired by @gmike11's work in #481, but that seems to
lump a bunch of changes together. This also solves #398.

Where #481 manipulated the XML request body to address that the
`<platformCore:nullFieldList>` needs to always use the `platformCore`
namespace, as opposed to the record's normal namespace (ie `transSale`
for invoice) as happens for any other field, I tried to address that on
the `#to_record` side.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -2,6 +2,8 @@
 
 ### Added
 
+* Add NullFieldList record (to credit memos and invoices) (#529)
+
 ### Fixed
 
 ## 0.8.11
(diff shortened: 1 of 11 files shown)


iloveitaly:
Awesome change, as always!

PR #530, opened 2022-04-12
Add `get_deleted` action to item records

cgunther:
I also added some documentation about how to use the action.

The documentation reveals a few rough parts around this action. Notably:

The response you get back is very raw, and not a typical record. Given
how little data is returned, I'm not sure if it'd be better to
instantiate a full record for consistency, or if there should be a
new `DeletedRecord`-style record, with a field for `deleted_date`
and a record_ref for `record`, making it a little easier to work with.

You have to paginate yourself. Eventually it'd be nice if something like
`SearchResult#results_in_batches` were offered to do it for you, though
that'd be a breaking change to the return value of the action currently.

The record class you call `get_deleted` on has no influence on the
request actually performed, so without any criteria, you'll get records
back that are totally unrelated to the record class it was called upon.
This goes beyond how search works where an inventory item search returns
all item types, this includes even non-items that were deleted. Perhaps
it'd be less surprising if the gem set a type criteria for you based on
the calling record type, though perhaps there's some value in the
current behavior of wanting to get all deleted records regardless of
type?

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -3,6 +3,7 @@
 ### Added
 
 * Add NullFieldList record (to credit memos and invoices) (#529)
+* Add `get_deleted` action to item records (#530)
 
 ### Fixed
 
--- a/README.md
+++ b/README.md
@@ -607,6 +607,53 @@ deposit.payment = 20
 deposit.add
 ```
 
+## Getting Deleted Records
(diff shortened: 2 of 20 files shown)


iloveitaly:
Thank you!

PR #532, opened 2022-04-15
gross_amt field in cash refund item

fabiensebban:
Add gross_amt field in cash refund item


yossix:
Good idea ! i need it too 8

PR #535, opened 2022-05-06
Update customer record (update fields, add search only fields)

cgunther:
I needed the `onCreditHold` field, which is only available via searches, so figured I'd add all the search only fields.

Before adding all the search only fields, figured it was wise to update all the regular fields since search only fields often seem to be a superset of the regular fields. A few fields were corrected to be `record_refs` and a few were removed as they require special sublist classes that aren't yet implemented.

There is a breaking change here renaming `CustomerSubscriptionsList` to `SubscriptionsList` and `CustomerSubscription` to `Subscription`. Looking at the schema for [Customer](https://system.netsuite.com/help/helpcenter/en_US/srbrowser/Browser2022_1/schema/record/customer.html), you'll see the type for `subscriptionsList` is `SubscriptionsList` (no Customer prefix). This is true as far back as [2014.1](https://system.netsuite.com/help/helpcenter/en_US/srbrowser/Browser2014_1/schema/record/customer.html) too. This break isn't strictly required, it could be removed from this PR, but seemed best to be consistent with NetSuite's record naming.


iloveitaly:
I think 2014.1 isn't supported, so I'm fine merging this in. Thanks again for all of the great PRs1

PR #539, opened 2022-05-17
Add get_item_availability operation

wkrsz:
Add get_item_availability operation that fetches ItemAvailability records for a list of InventoryItem refs.

https://docs.oracle.com/en/cloud/saas/netsuite/ns-online-help/section_N3498308.html

Is NetSuite::Records::ItemAvailability class the right place for this operation? Or should it go to Actions?

### DIFF
--- a/lib/netsuite/records/item_availability.rb
+++ b/lib/netsuite/records/item_availability.rb
@@ -16,6 +16,28 @@ module NetSuite
       field :quantity_committed
       field :quantity_available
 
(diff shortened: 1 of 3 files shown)


wkrsz:
It accepts different item record types: inventory item, gift certificate item etc. The output is always ItemAvailability record.

iloveitaly:
Perfect, this is a great approach then!

iloveitaly:
@wkrsz mind adding a changelog entry in a separate PR?

PR #541, opened 2022-05-23
Add created_from_ship_group field to ItemFulfillment

wkrsz:
CleanShot 2022-05-23 at 14 15 06@2x


iloveitaly:
Surprised this field isn't a record ref!

PR #543, opened 2022-05-23
Handle empty result in get_select_value

wkrsz:
This fixes an internal exception when parsing empty response to `get_select_value`.

This particular response (with a warning) was recorded on a NetSuite account organized into subsidiaries. To get list of valid ship methods we'd need to also pass a customer filter from which NetSuite infers subsidiary and available ship methods.

```
NoMethodError:
  private method `select' called for nil:NilClass
# ./lib/netsuite/support/attributes.rb:14:in `initialize_from_attributes_hash'
# ./lib/netsuite/support/sublist.rb:34:in `initialize'
# ./lib/netsuite/actions/get_select_value.rb:52:in `new'
# ./lib/netsuite/actions/get_select_value.rb:52:in `get_select_value'
# ./spec/netsuite/actions/get_select_value_spec.rb:9:in `block (3 levels) in <top (required)>'
# ./spec/netsuite/actions/get_select_value_spec.rb:38:in `block (4 levels) in <top (required)>
```

### DIFF
--- a/lib/netsuite/support/sublist.rb
+++ b/lib/netsuite/support/sublist.rb
@@ -31,7 +31,7 @@ module NetSuite
       end
 
       def initialize(attributes = {})
-        initialize_from_attributes_hash(attributes)
+        initialize_from_attributes_hash(attributes || {})
       end
 
       def to_record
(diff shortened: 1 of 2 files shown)


iloveitaly:
I think we should probably return a more structured error here. There has been some discussion about this in the past.

However, this is an improvement over the currently-terrible exception that is thrown, so let's merge this in!

wkrsz:
Thanks! I figured the test with XML fixture will aid in testing the improved error handling.

PR #548, opened 2022-06-14
Optionally pass in `proxy` parameter to Savon

dbecker-stripe:
Savon will call `to_s` on any `proxy` value passed in so it will treat `nil` like an empty string instead of not using it. This causes requests to fail with an invalid proxy. This change will not pass in a value for `proxy` if the attribute is `nil`.

This change also updates history with addition of the proxy attribute.

A [fix was added to version 2.12.0 of savon](https://github.com/savonrb/savon/blob/master/CHANGELOG.md#2120-2018-01-16) but this changes allows netsuite to be compatible with earlier versions of savon.

### DIFF
--- a/lib/netsuite/configuration.rb
+++ b/lib/netsuite/configuration.rb
@@ -15,8 +15,6 @@ module NetSuite
     end
 
     def connection(params={}, credentials={})
-      proxy_params = proxy ? {proxy: proxy} : {}
-      
       client = Savon.client({
         wsdl: cached_wsdl || wsdl,
         endpoint: endpoint,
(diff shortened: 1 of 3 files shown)


lib/netsuite/configuration.rb
```diff
@@ -15,6 +15,8 @@ def attributes
     end
 
     def connection(params={}, credentials={})
+      proxy_params = proxy ? {proxy: proxy} : {}
```
iloveitaly, line 18: @dbecker-stripe  Can you add a test for setting the proxy config to the test suite?

iloveitaly (changes requested):

lib/netsuite/configuration.rb
dbecker-stripe: Added testing. I'm not sure they will catch the bug given that it requires an older version of savon but the tests run against the latest version. If there is a change in savon behavior these tests will catch it.

PR #549, opened 2022-06-16
Making gem thread safe

andrewdicken-stripe:
Making gem thread safe by wrapping instance variables in `Thread.current`. All tests are passing.

### DIFF
--- a/lib/netsuite/configuration.rb
+++ b/lib/netsuite/configuration.rb
@@ -51,7 +51,7 @@ module NetSuite
     end
 
     def wsdl_cache
-      @wsdl_cache ||= {}
+      Thread.current[:netsuite_gem_wsdl_cache] ||= {}
     end
 
     def clear_wsdl_cache
--- a/lib/netsuite/utilities.rb
+++ b/lib/netsuite/utilities.rb
@@ -203,7 +203,6 @@ module NetSuite
       opts[:external_id] ||= false
 
       if opts[:cache]
-        netsuite_get_record_cache ||= {}
         netsuite_get_record_cache[record_klass.to_s] ||= {}
 
         if netsuite_get_record_cache[record_klass.to_s].has_key?(id.to_i)
@@ -228,7 +227,7 @@ module NetSuite
       rescue ::NetSuite::RecordNotFound
         # log.warn("record not found", ns_record_type: record_klass.name, ns_record_id: id)
         if opts[:cache]
-          @netsuite_get_record_cache[record_klass.to_s][id.to_i] = nil
+          netsuite_get_record_cache[record_klass.to_s][id.to_i] = nil
         end
 
         return nil
@@ -243,8 +242,6 @@ module NetSuite
       # FIXME: Records that have the same name but different types will break
       # the cache
       names.each do |name|
-        netsuite_find_record_cache ||= {}
-
         if netsuite_find_record_cache.has_key?(name)
           return netsuite_find_record_cache[name]
         end
--- a/spec/netsuite/configuration_spec.rb
+++ b/spec/netsuite/configuration_spec.rb
@@ -14,6 +14,20 @@ describe NetSuite::Configuration do
       config.reset!
       expect(config.attributes).to be_empty
     end
+
+    it 'ensures that attributes are not shared between threads' do
+      config.attributes[:blah] = 'something'
+      expect(config.attributes[:blah]).to eq('something')
+
+      thread = Thread.new {
+        config.attributes[:blah] = 'something_else'
+        expect(config.attributes[:blah]).to eq('something_else')
+      }
+
+      thread.join
+
+      expect(config.attributes[:blah]).to eq('something')
+    end
   end
 
   describe '#filters' do


iloveitaly:
@andrewdicken-stripe can we add a test that creates a thread, modifies the NS config, and then ensures that the NS config on the main thread remains unchanged?

iloveitaly:
@andrewdicken-stripe can you add a changelog entry in a separate PR?

cgunther:
I think there's a bit of a flaw in this approach, at least for my setup.

After upgrading to a version including this change, I started to see the following errors:
```
Savon::HTTPError: HTTP error (302): <html><head><title>302 Moved Temporarily</title></head>
<body bgcolor="#FFFFFF">
<p>This document you requested has moved temporarily.</p>
<p>It's now at <a href="https://webservices.netsuite.com/pages/portal/page_not_found.jsp?internal=F">https://webservices.netsuite.com/pages/portal/page_not_found.jsp?internal=F</a>.</p>
</body></html>
```

My setup is a Rails app where NetSuite is configured in an initializer (only dealing with a single NetSuite account), then most requests are happening in Sidekiq jobs.

I think what's happening is that NetSuite is configured in the main Sidekiq server thread as it loads the Rails environment and runs the initializers, but then each Sidekiq job runs in it's own separate thread, and the `Thread.current` values are not copied from the main Sidekiq server thread to the individual job thread. Therefore I think that leaves each Sidekiq job running as if NetSuite were never configured at all.

The approach this PR takes likely works if the configuration is happening in the same thread that the request will happen, but not if those happen in separate threads.

I'm not super familiar with threads, but it seems like we need something so after NetSuite is configured, any new threads copy the existing configuration, but then any modifications to that configuration in the new thread should remain local to that thread.

Here's a quick simplified reproduction:
```ruby
require 'netsuite'

NetSuite.configure do
  reset!

  # Configure me...
end

NetSuite::Records::CustomerCategory.search.results

Thread.new do
  NetSuite::Records::CustomerCategory.search.results
end.join
```

If you put that in the gem's root path, you can run it with `bundle exec ruby filename.rb`.

The first request (main thread) runs fine, but then the second request (new thread) errors with `(soapenv:Server.userException) Invalid account number. (Savon::SOAPFault)`. The request XML on the second request is:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<env:Envelope xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:platformMsgs="urn:messages_2016_2.platform.webservices.netsuite.com" xmlns:env="http://schemas.xmlsoap.org/soap/envelope/" xmlns:platformCore="urn:core_2016_2.platform.webservices.netsuite.com" xmlns:platformCommon="urn:common_2016_2.platform.webservices.netsuite.com" xmlns:listRel="urn:relationships_2016_2.lists.webservices.netsuite.com" xmlns:tranSales="urn:sales_2016_2.transactions.webservices.netsuite.com" xmlns:tranPurch="urn:purchases_2016_2.transactions.webservices.netsuite.com" xmlns:actSched="urn:scheduling_2016_2.activities.webservices.netsuite.com" xmlns:setupCustom="urn:customization_2016_2.setup.webservices.netsuite.com" xmlns:listAcct="urn:accounting_2016_2.lists.webservices.netsuite.com" xmlns:tranBank="urn:bank_2016_2.transactions.webservices.netsuite.com" xmlns:tranCust="urn:customers_2016_2.transactions.webservices.netsuite.com" xmlns:tranEmp="urn:employees_2016_2.transactions.webservices.netsuite.com" xmlns:tranInvt="urn:inventory_2016_2.transactions.webservices.netsuite.com" xmlns:listSupport="urn:support_2016_2.lists.webservices.netsuite.com" xmlns:tranGeneral="urn:general_2016_2.transactions.webservices.netsuite.com" xmlns:commGeneral="urn:communication_2016_2.general.webservices.netsuite.com" xmlns:listMkt="urn:marketing_2016_2.lists.webservices.netsuite.com" xmlns:listWebsite="urn:website_2016_2.lists.webservices.netsuite.com" xmlns:fileCabinet="urn:filecabinet_2016_2.documents.webservices.netsuite.com" xmlns:listEmp="urn:employees_2016_2.lists.webservices.netsuite.com">
  <env:Header>
    <platformMsgs:passport>
      <platformCore:account/>
      <platformCore:email xsi:nil="true">***FILTERED***</platformCore:email>
      <platformCore:password xsi:nil="true">***FILTERED***</platformCore:password>
      <platformCore:role internalId="3"/>
    </platformMsgs:passport>
    <platformMsgs:SearchPreferences/>
  </env:Header>
  <env:Body>
    <platformMsgs:search>
      <platformMsgs:searchRecord xsi:type="listAcct:CustomerCategorySearch"/>
    </platformMsgs:search>
  </env:Body>
</env:Envelope>
```
Note the account element is empty.

If I run that same script against 64b6199, it works just fine.

iloveitaly:
@cgunther In your case, do you not initialize the Netsuite configuration within each thread? i.e.

```ruby
class Worker
  include Sidekiq::Worker

  def perform
    init_netsuite_client
  end
end
```

I'm guessing you have some sort of `config/initializers/netsuite.rb` which configures everything globally?

Your scenario makes sense. I'm open to some sort of `single_tenant = true` option which copies the NS config across threads!

cgunther:
Correct, `config/initializers/netsuite.rb`, as opposed to `init_netsuite_client` in each worker, as the latter would be redundant for me being single tenant. The majority of my requests are in workers, but a few happen from controllers, and the initializer is one place to configure the gem regardless of where I'm going to use it. Most other API clients I'm able to configure in the same way, via initializer, rather than repeat per use.

Should `single_tenant` be an option, or should the copying behavior just be automatic? Speculating on a multi-tenant use case, I'd imagine you may still want to set something like `read_timeout` globally in something like an initializer, then per-thread configure the gem specific to the account you're accessing, so I don't think copying in multi-tenant as a default starting point would be a bad thing.

That said, any tips on how to actually accomplish this? I'm no expert on threading and a quick search didn't surface any hooks in Ruby for detecting when a new thread has been created for us to copy those variables. I imagine we don't necessarily want to tailor this specifically to Sidekiq and how it creates threads for jobs as we'd want to affect threads created for any reason (Sidekiq, Puma, manually, etc.).

iloveitaly:
@cgunther I think we should do something like:

* Add a `multi_tenant!` configuration option. Default behavior should be single tenant
* If multi_tenant is enabled, use `Thread.main` for config storage, otherwise use `Thread.current` ([This pattern](https://github.com/FundingCircle/virsandra/blob/69702471f2507bd89270d1c7bbee3d5390a2af84/lib/virsandra.rb#L13-L22) looks like a clean way to handle this.)

What do you think?

PR #554, opened 2022-07-18
update-default-api-version

Timothyjb:
I began getting the following error
```
(soapenv:Server.userException) SuiteTalk endpoint version 2015_1 is no longer supported
```
when not explicitly setting the api_version.

I am just updating the default version to 2016_2 (though it may make more sense to default to a later version)


iloveitaly:
Nice! Thank you.

PR #556, opened 2022-08-18
Add `Configuration#multi_tenant!` for opting into multi-tentant support where configuration is per-thread

cgunther:
This is a follow up to #549 to improve usage in a single tenant scenario where you might configure NetSuite on the main thread (ie. via Rails initializer), then make requests in another thread (ie. Sidekiq jobs).

By default we assume single-tenant usage and the configuration is shared across all threads. If you opt into multi-tenant usage, each child thread starts with the main thread's configuration, but can then be further configured without altering other threads.

Rather than mixing `Thread.current` and global instance variables, I'm using `Thread.main` for the global configuration.

Child threads start with a copy of the main thread's attributes, figuring this still allows you to set some things globally on the main thread (ie. timeouts), then per-tenant things on the child threads (ie. account).

Whether we're multi-tentant or not is always stored on the main thread, not as an attribute, as we want/need that to be global.

One possible concern is what happens if you set some configuration before calling `multi_tenant!`, particularly when called from a child thread. Whatever configuration you set before `multi_tenant!` would be stored on the main thread (thinking we're single-tenant), not the child thread, which could be surprising. Maybe we need a safety check so it'll error if you call `multi_tenant!` after attributes have already been set, forcing you to specify `multi_tentant!` first?

`reset!` also resets whether it's multi-tenant, taking the stance that it's a full reset, not just resetting the thread.

I didn't touch caches yet, wanted to make sure we're on the right path first. I'll need to mention this in the README too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,6 +159,30 @@ NetSuite.configure do
 end
 ```
 
+### Multi-Tenancy
+
+If you're interacting with multiple NetSuite accounts, each in separate threads, you can enable multi-tenancy to prevent your configuration and caches from being shared between threads.
+
+From your main thread, you'd want to enable multi-tenancy:
+
+```ruby
+NetSuite.configure do
+  multi_tentant!
+end
+```
+
+Then in each child thread, you'd perform any configuration specific to the NetSuite account you're interacting with for that thread, all of which will be specific to that thread only:
+
+```ruby
+NetSuite.configure do
+  reset!
+
+  account ENV['NETSUITE_ACCOUNT']
+
+  # The rest of your usual configuration...
+end
+```
+
 # Usage
 
 ## CRUD Operations
--- a/README.md
+++ b/README.md
@@ -171,6 +171,8 @@ NetSuite.configure do
 end
 ```
 
+Note that `multi_tenant!` is a special configuration option which is _not_ effected by `reset!`.
+
 Then in each child thread, you'd perform any configuration specific to the NetSuite account you're interacting with for that thread, all of which will be specific to that thread only:
 
 ```ruby
--- a/lib/netsuite/configuration.rb
+++ b/lib/netsuite/configuration.rb
@@ -11,7 +11,11 @@ module NetSuite
     end
 
     def attributes
-      Thread.current[:netsuite_gem_attributes] ||= {}
+      if multi_tenant?
+        Thread.current[:netsuite_gem_attributes] ||= {}
+      else
+        @attributes ||= {}
+      end
     end
 
     def connection(params={}, credentials={})
@@ -51,11 +55,19 @@ module NetSuite
     end
 
     def wsdl_cache
-      Thread.current[:netsuite_gem_wsdl_cache] ||= {}
+      if multi_tenant?
+        Thread.current[:netsuite_gem_wsdl_cache] ||= {}
+      else
+        @wsdl_cache ||= {}
+      end
     end
 
     def clear_wsdl_cache
-      Thread.current[:netsuite_gem_wsdl_cache] = {}
+      if multi_tenant?
+        Thread.current[:netsuite_gem_wsdl_cache] = {}
+      else
+        @wsdl_cache = {}
+      end
     end
 
     def cached_wsdl
@@ -406,6 +418,14 @@ module NetSuite
       else
         attributes[:proxy]
       end
+    end
+
+    def multi_tenant!
+      @multi_tenant = true
+    end
+
+    def multi_tenant?
+      @multi_tenant
     end
   end
 end
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -2,6 +2,8 @@
 
 
 ### Added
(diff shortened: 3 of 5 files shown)


lib/netsuite/configuration.rb
```diff
@@ -407,5 +412,9 @@ def proxy(proxy = nil)
         attributes[:proxy]
       end
     end
+
+    def multi_tenant!
+      Thread.main[:netsuite_gem_multi_tenant] = true
```
iloveitaly, line 417: This is a global var so we can use `@multi_tenant` to store this value. Using a instance variable is better here: consistent with the rest of the config and leaks less configuration into global/thread-global state.

lib/netsuite/configuration.rb
```diff
@@ -7,11 +7,16 @@ def reset!
...
       attributes.clear
     end
 
     def attributes
-      Thread.current[:netsuite_gem_attributes] ||= {}
+      if Thread.main[:netsuite_gem_multi_tenant]
+        Thread.current[:netsuite_gem_attributes] ||= (Thread.main[:netsuite_gem_attributes] || {}).dup
+      else
+        Thread.main[:netsuite_gem_attributes] ||= {}
+      end
```
iloveitaly, line 19: My thought here is something like:

```suggestion
      if @multi_tenant
        Thread.current[:netsuite_gem_attributes] ||= {}
      else
        @attributes
      end
```

i.e. we don't want to copy config if we are multi-tenant, it's up to each process to set the NS config before processing.

iloveitaly:
Thank you for the great contribution! Left some comments.

cgunther:
Thanks for the feedback, took that into account.

I followed the same pattern with attributes to apply to the caches that were changed in #549.

Also documented the configuration in the README.

I'm still unsure of whether `Configuration#reset!` should reset multi-tenancy or not. All the examples in the README start with a `reset!`, which seems unnecessary, at least in a single-tenant scenario, which is the extent of my experience. Trying to speculate on the multi-tenant scenario, if you're going to start your configuration in each child thread with a `reset!`, then you'd also need to follow it with a `multi_tenant!` call since the reset currently resets back to single-tenant mode. Broadly, I think we need to decide if multi-tenancy is a configuration (just like account, api_version, etc.), such that it's resettable, or is it a new higher-level concept that is not resettable (ie. `NetSuite.multi_tenant!`), and thus the configuration is resettable, but that doesn't change the multi-tenancy.

iloveitaly:
> Broadly, I think we need to decide if multi-tenancy is a configuration (just like account, api_version, etc.), such that it's resettable, or is it a new higher-level concept that is not resettable (ie. `NetSuite.multi_tenant!`), and thus the configuration is resettable, but that doesn't change the multi-tenancy.

I think this is the right framing of the problem. My thought is `multi_tenant!` should be a separate concept (maybe with it's own instance var?) which is not effected by `reset!`.

That is a bit weird, but having it effected by `reset!` would be worse.

Wdyt?

cgunther:
I tweaked things so `reset!` no longer resets multi-tenancy (it's stored as it's own instance variable, not within the attributes hash) and updated the README to reflect that you'd probably configure once in your main thread to enable multi-tenancy, then configure again in each child thread to set your other attributes.

The test gets a little wonky as we have to use `instance_variable_set` to reset multi-tenancy after enabling it to ensure it doesn't leak to other tests, as we don't publicly offer a way to disable multi-tenancy.

It slightly irks me that `#multi_tenant!` lives in `Configuration` since it's really at a higher level, but there's a smidge of precedent for non-attribute methods in this class (ie. `#connection`, `#wsdl_cache`, etc.) so maybe it's ok for now.

Ultimately I'd really like feedback from someone with a multi-tenant setup, as my experience is only with single-tenant setups, so I'm just speculating on how a multi-tenant setup might look. My motivation with this PR was to address that #549 broke my single-tenant set up.

README.md
```diff
@@ -159,6 +159,30 @@ NetSuite.configure do
...
+
+From your main thread, you'd want to enable multi-tenancy:
+
+```ruby
+NetSuite.configure do
+  multi_tentant!
+end
+```
+
+Then in each child thread, you'd perform any configuration specific to the NetSuite account you're interacting with for that thread, all of which will be specific to that thread only:
```
iloveitaly, line 174:
```suggestion
Note that `multi_tenant!` is a special configuration option which is _not_ effected by `reset!`.

Then in each child thread, you'd perform any configuration specific to the NetSuite account you're interacting with for that thread, all of which will be specific to that thread only:
```

iloveitaly:
@cgunther Awesome work! I have a multi-tenant setup. Going to merge this in and run it against our internal test suite.

cgunther:
Thanks for the merge.

Things seem to work well for my single-tenant setup. Assuming they work well for your multi-tenant setup, it might be wise to release a new version. I suspect the current version (v0.9.0) would be broken for anyone with a single-tenant setup, but using threads (ie. puma, sidekiq) and configuring outside those child threads as a result of #549.

iloveitaly:
@cgunther definitely! I'll get a release out soon :)

PR #558, opened 2022-08-22
Locking Savon to < 2.13 to prevent an error making queries

avanrielly:
Making a request with the NetSuite gem and Savon 2.13.0 caused the following error:

```
<faultstring>org.xml.sax.SAXParseException; lineNumber: 1; columnNumber: 435; The prefix "xmlns" cannot be bound to any namespace explicitly; neither can the namespace for "xmlns" be bound to any prefix explicitly.
```

Therefore, this PR locks the Savon gem to the last versions that work.

If I had to guess at the PR that caused the issue, I would suggest it is [this](https://github.com/savonrb/savon/pull/943) one.

### DIFF
--- a/netsuite.gemspec
+++ b/netsuite.gemspec
@@ -20,7 +20,7 @@ Gem::Specification.new do |gem|
   gem.metadata['mailing_list_uri'] = 'http://opensuite-slackin.herokuapp.com'
   gem.metadata['rubygems_mfa_required'] = 'true'
 
-  gem.add_dependency 'savon', '>= 2.3.0'
+  gem.add_dependency 'savon', '>= 2.3.0', '< 2.13'
 
   gem.add_development_dependency 'rspec', '~> 3.11.0'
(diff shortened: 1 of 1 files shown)


iloveitaly:
@avanrielly great! Going to merge, but can you submit another PR with a changelog entry?

avanrielly:
@iloveitaly Sorry, I did not even realize this was part of the process.  I have made a follow-up PR [here](https://github.com/NetSweet/netsuite/pull/560)

cgunther:
We may want to revert this (and #560) as the underlying issue in Savon was fixed and released as v2.13.1, and I confirmed the new version works correctly for me:
https://github.com/savonrb/savon/pull/977

v2.13.0 will continue to be broken, but restricting anyone from using the newer v2.13.1 (or future versions) doesn't seem ideal.

iloveitaly:
@cgunther great idea! I think we can ban specific versions using `!=`. More info here: https://guides.rubygems.org/patterns/#pessimistic-version-constraint

PR #564, opened 2022-09-16
Save response when error occur

fabiensebban:
This modification will allow people get the NetSuite's response when an error occurs on an `initialize` request.

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,1 +1,1 @@
-github: [NetSweet]
+github: [iloveitaly]


iloveitaly:
how would this work? since an exception is raised the user would not get a reference, right?

fabiensebban:
Wow! My apologies. I made the PR too fast.
I've been looking for return NetSuite's API response whenever an error occurs during a request. Sometimes there is helpful information in the response. Now I face another problem: I don't know how to get the response to give it to the InitializationError exception I just created in my previous commit. Do you know how can I get the response so I can handle the initialization exception like follow?
```
raise NetSuite::InitializationError.new("#{self}.initialize with #{object} failed.", @response_hash)
```

PR #565, opened 2022-09-29
Add search only fields to `CreditMemo`

cgunther:
Both `Invoice` and `CreditMemo` records (as well as many other records) use the same search fields for the returned basic results, `TransactionSearchRowBasic`, so rather than repeat the fields for each record, extract them to a common module that can be shared between the respective records.

Since the search module may cover multiple records, it's fields will overlap with some of the standard fields on each record, so defining search only fields is now a no-op if the search only field already exists on the record as a standard field (rather than erroring). Previously for invoices, we defined search only fields as only the fields that didn't otherwise exist as standard fields, but now we need to include those standard fields potentially as search only as well as they may not be standard fields on another record that uses the same module. For this reason, it's necessary to include the search module after defining all the standard fields, otherwise we'll error if we try to define a standard field if it's already been defined as a search only field.

At this point, I only have use for searching invoices and credit memos, so the module is only included on those records, but it's the same search fields that could eventually be included on `AssemblyBuild`, `BinTransfer`, `CashSale`, etc. records, all of which generally seem to fall under "transaction".

---

Alternatively, I'm conflicted whether `search` should continue to be an action on the individual records, as it is now, or whether it'd be better off as a separate class, better mirroring how it works internally at NetSuite.

For example, `Invoice.search` will return all transaction types (ie. invoice, credit memo, etc.) unless you add a criteria on `type` to limit to just invoices. Similarly, `InventoryItem.search` will return all item types (ie. inventory, non-inventory, service, etc.) unless you add a criteria on `type`. And `Customer.search` will return all customer stages (ie. lead, prospect, customer) unless you add a criteria on `stage`. There's probably more examples that I haven't come across yet. This regularly surprises me.

Furthermore, `Invoice.search` may include a credit memo in it's results, but the instance of that record will be of type `NetSuite::Records::Invoice`, since that's the record class you searched on, as opposed to `NetSuite::Records::CreditMemo`, which is misleading.

Taking it a step further, `Invoice.search` may include the individual lines of each invoice as well, unless you set your criteria to mainline only, and now each of those lines is an instance of `NetSuite::Records::Invoice`, but that's a bit of a stretch.

One simple thought is that when you call `Invoice.search`, we could automatically add the criteria for `type`. This would align with the NetSuite UI where if you're looking at the list of invoices and click Search in the top-right, it defaults the resulting search to "Type: Invoice". Then we can guarantee all results from `Invoice.search` would actually be invoices and having them be instances of `NetSuite::Records::Invoice` would always be appropriate (though less so for the individual lines, unless we defaulted to mainline only, but that seems like a step too far).
Screen Shot 2022-09-29 at 2 54 46 PM

The downside there being you couldn't search across types, if you were so inclined, unless you searched on each separate class (ie invoice, credit memo, etc.) then combined them yourself. Perhaps a solution there is to have a class like `TransactionSearch` that you use, so now it's clear you're searching transactions generally, and if you want to narrow it down by type, you have to provide the criteria. Then the results could be instances of something like `TransactionSearchRow`, where we'd define the fields that could be returned aligning with NetSuite's [TransactionSearchRowBasic](https://system.netsuite.com/help/helpcenter/en_US/srbrowser/Browser2022_1/schema/search/transactionsearchrowbasic.html?mode=package). The downside there, however being given an instance of a result, it'd be on the developer to cast that into an instance of the proper record (ie. `Invoice`) if they needed to do some updates, or perhaps the row instance offers some helper method to convert to the record. Maybe `Invoice.search` could still exist, but just proxy to `TransactionSearch` with the appropriate criteria for type?

Those are just my thoughts on using NetSuite and this gem after a few years.


cgunther:
Updated to add search only fields to `PurchaseOrder` as well.

cgunther:
Updated to add the search only fields to `WorkOrder` as well.

PR #568, opened 2022-11-01
[WIP] Create abstract action

andrewdicken-stripe:


andrewdicken-stripe:
Question for code reviewer: `lib/netsuite/actions/login.rb` is within the actions folder but does not follow the normal pattern of having an inner `Support` module and seems to be structured fairly different than the other actions. Is it worth implementing the abstract action pattern for that file or should we just leave that one out of this change?

lib/netsuite/actions/add.rb
```diff
@@ -1,7 +1,9 @@
 # https://system.netsuite.com/help/helpcenter/en_US/Output/Help/SuiteCloudCustomizationScriptingWebServices/SuiteTalkWebServices/add.html
+require_relative 'abstract_action'
```
iloveitaly, line 2: I don't think we need these, I'm pretty sure we can `autoload` them in `lib/netsuite`

lib/netsuite/actions/abstract_action.rb
```diff
@@ -0,0 +1,32 @@
...
+
+      def initialize
+        raise NotImplementedError, 'Not implemented on abstract class'
+      end
+
+      def request_body
+        raise NotImplementedError, 'Not implemented on abstract class'
+      end
+
+      def request_options
```
iloveitaly, line 22: is there a reason we wouldn't default to `{}` and omit these dup methods on the specific actions?

lib/netsuite/actions/abstract_action.rb
```diff
@@ -0,0 +1,32 @@
+module NetSuite
+  module Actions
+    class AbstractAction
+      def request(credentials={})
```
iloveitaly, line 4: can you link to this PR in a comment to help explain why this pattern is being used?

iloveitaly (changes requested):

lib/netsuite/actions/add.rb
andrewdicken-stripe: Ahhh I didn't see that I could do that. Much cleaner.

lib/netsuite/actions/abstract_action.rb
andrewdicken-stripe: I added a link to this PR to my WIP PR where I'm implementing the monkey patch for `AbstractAction`. Is that what you were referring to? https://git.corp.stripe.com/stripe-internal/pay-server/pull/596685/files#r2504037

iloveitaly (approved):

PR #574, opened 2023-03-16
dont crash if suitetalk returns empty <platformCore:searchRowList/>

ericcj:
The client should just yield the blank page returned instead of crashing, but to ever receive this seems like a bug in suitetalk since we're using results_in_batches but on the final page sometimes a successful result is returned with no rows, probably because the underlying data no longer matches the search since our pagination takes an hour or so and we're running it across day boundaries, but you'd think search_id would represent an isolated view of the data we're paginating.  we haven't captured the log to see if total_records changes, but obviously total_pages doesn't since it successfully accepts page_index but just doesn't return any rows.

### DIFF
--- a/lib/netsuite/support/search_result.rb
+++ b/lib/netsuite/support/search_result.rb
@@ -48,7 +48,8 @@ module NetSuite
             end
           elsif response.body.has_key? :search_row_list
             # advanced search results
-            record_list = response.body[:search_row_list][:search_row]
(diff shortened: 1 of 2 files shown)


iloveitaly:
@ericcj can you include a test for this with a XML fixture? Thanks for the contribution!

ericcj:
@iloveitaly done

iloveitaly:
Awesome, thank you!

Issue #578, opened 2023-04-19
Recent proxy change caused proxy behavior to break


andrewdicken-stripe:
[This](https://github.com/NetSweet/netsuite/pull/548) recent proxy change released in version 0.9.3 does not appear to be working as intended. When we deployed this change, all of our requests were not going through our defined proxy. Since this appears like a regression and does not allow proxies to be used with the NetSuite gem, can we revert that change (will put up a new PR shortly) and cut a new gem version? This will fix the broken behavior for users who make NetSuite calls via proxies.

PR #579, opened 2023-04-19
Revert proxy change

andrewdicken-stripe:
Reverts recent proxy related change, explained further in issue here: https://github.com/NetSweet/netsuite/issues/578. I'm not exactly sure why the proxy change breaks things but this fix should revert the behavior to how it was before.

### DIFF
--- a/spec/netsuite/configuration_spec.rb
+++ b/spec/netsuite/configuration_spec.rb
@@ -519,21 +519,13 @@ describe NetSuite::Configuration do
       expect(config.proxy).to be_nil
     end
 
-    it 'does not pass in nil proxy to savon' do
-      connection = config.connection
-
-      expect(connection.globals.include?(:proxy)).to eql(false)
-    end
-
     it 'can be set with proxy=' do
       config.proxy = "https://my-proxy"
 
       expect(config.proxy).to eql("https://my-proxy")
 
       # ensure no exception is raised
-      connection = config.connection
-
-      expect(connection.globals.include?(:proxy)).to eql(true)
+      config.connection
     end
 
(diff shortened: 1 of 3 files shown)


spec/netsuite/configuration_spec.rb
```diff
@@ -519,21 +519,13 @@
       expect(config.proxy).to be_nil
     end
 
-    it 'does not pass in nil proxy to savon' do
```
iloveitaly, line 522: can we keep this test? this is an important case to cover

spec/netsuite/configuration_spec.rb
```diff
@@ -519,21 +519,13 @@
...
     it 'can be set with proxy=' do
       config.proxy = "https://my-proxy"
 
       expect(config.proxy).to eql("https://my-proxy")
 
       # ensure no exception is raised
-      connection = config.connection
-
-      expect(connection.globals.include?(:proxy)).to eql(true)
+      config.connection
```
iloveitaly, line 528: I think we'd want to keep this case too

iloveitaly (changes requested):
Mmm, it's not clear why this caused an issue. Can we keep the additional test cases here and maybe add some more? Then I'm happy to revert.

iloveitaly:
@andrewdicken-stripe friendly ping here! Would love to get this change in.

andrewdicken-stripe:
@iloveitaly I made the change you requested, but for some reason the builds started failing. Not sure if there's an issue with the test suite because I had it working before and the changes that I made shouldn't have caused tall the issues I'm seeing.

PR #583, opened 2023-06-22
Netsuite.v0.9.3 with BOM

george-pmk:


iloveitaly:
@george-pmk could you:

1. Remove the version change from this?
2. Split this into a couple of smaller PRs? Will be easier to review that way

PR #587, opened 2023-08-11
Accept write_timeout configuration, and supply it along to savoy

nevinera:
The default timeout for reading from netsuite in the gem's configuration is 60 seconds, and the 'open' timeout can be set, but the 'write' timeout was not settable, and defaulted to [0.25 seconds](https://www.rubydoc.info/github/httprb/http/HTTP/Timeout/PerOperation).

Allow the write-timeout to be configured, and supply it to Savoy alongside the read and open timeouts.

### DIFF
--- a/lib/netsuite/configuration.rb
+++ b/lib/netsuite/configuration.rb
@@ -24,6 +24,7 @@ module NetSuite
         endpoint: endpoint,
         read_timeout: read_timeout,
         open_timeout: open_timeout,
+        write_timeout: write_timeout,
         namespaces: namespaces,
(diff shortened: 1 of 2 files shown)


iloveitaly:
looks great, thanks!

nevinera:
@iloveitaly - definitely, I didn't realize how far back the support reached. Looks like `write_timeout` was [added in 2.13.0,](https://github.com/savonrb/savon/blob/master/CHANGELOG.md#2130-2022-08-03) I'll include a version check and enable the test-suite on my end.

PR #600, opened 2024-01-22
Return SOAP error context for failed initializations

bebanowski:
We want to see the see correct error information when an initialization action fails. The existing gem behaviour is to return a generic failure message. Instead of this, we want the failure case to raise an InitializationError using the error message contained in the SOAP response.


bebanowski:
Created in error. This pull request was intended for my development fork of this repository.

PR #606, opened 2024-03-06
Handle string field names in `CustomFieldList#respond_to?`

cgunther:
Previously `respond_to?` assumed it was given a symbol argument, so it directly checked for that argument in the fields, which are force-keyed as symbols. If you passed it a string argument, it wouldn't find the corresponding field, impying it didn't respond to such a method.

Now, the argument is explicitly cast to a symbol before checking if such a field exists, supporting checking `respond_to?` with both argument types.


cgunther:
CI is failing because of an upstream change in Nori, which can be fixed by #607

PR #608, opened 2024-03-07
Define `respond_to?` on `RecordRef` to match `method_missing` behavior

cgunther:
`RecordRef`s often include a name, which allows access via `method_missing` by calling `record_ref.name`, however if no referenced record is assigned, you'll still get a `RecordRef` instance, it just wont have any fields, so `record_ref.name` raised a `NoMethodError`, but if you checked for the field first via `record_ref.respond_to?(:name)`, it'd be falsey regardless of whether the field existed or not.

Defining `respond_to?` in good practice alongside `method_missing`. Ideally, `respond_to_missing?` would be used instead, as that enables access to the method like `record_ref.method(:name)`, however I kept with `respond_to?` to match `CustomFieldList`.

`CustomRecordRef` also uses `method_missing`, however there's no specs and I'm not familiar with that record, so I left it alone.


cgunther:
CI is failing because of an upstream change in Nori, which can be fixed by https://github.com/NetSweet/netsuite/pull/607

PR #614, opened 2024-05-09
Handle single result in get_item_availability

wkrsz:
When get_item_availability returns single result, Savon doesn't wrap it in an array. This PR handles such response.


wkrsz:
I wanted to use `Array.wrap` but it's not available in Ruby 2.7 which is still supported.

